PR #5, opened 2018-11-13
NXP-26145: Policy is now part of the settings

bdelbosc:
The policy defines retries, fallback and batching properties for computation
that extends AbstractBatchComputation.


covolution:
:+1:

pierre-gautier:
👍

PR #35, opened 2018-12-12
NXP-26466: Improve logging for dataset upload

covolution:


nuxeo-ai-model/src/main/java/org/nuxeo/ai/bulk/DataSetUploadComputation.java
```diff
@@ -44,8 +48,12 @@
 
     private static final Log log = LogFactory.getLog(DataSetUploadComputation.class);
 
+    protected final Timer uploadTimer;
+
     public DataSetUploadComputation(String name) {
         super(name, 1, 0);
+        MetricRegistry registry = SharedMetricRegistries.getOrCreate(MetricsService.class.getName());
+        uploadTimer = registry.timer(MetricRegistry.name("nuxeo", "ai", "cloud", "upload"));
```
kevinleturc: If I remember well, metric stuff can be registered only once per instance. You probably need to remove them when computation shutdown + make it unique per computation (if several are running on same node)

nuxeo-ai-model/src/main/java/org/nuxeo/ai/bulk/DataSetUploadComputation.java
```diff
@@ -65,10 +73,15 @@ public void processRecord(ComputationContext context, String inputStreamName, Re
                                 if (document != null) {
                                     CloudClient client = Framework.getService(CloudClient.class);
                                     if (client.isAvailable()) {
-                                        client.uploadDataset(document);
+                                        log.info(String.format("Uploading dataset to cloud for command %s", cmd.getId()));
```
kevinleturc: You could use log4j2 loggers to avoid `format` computation if log level is not info.

nuxeo-ai-model/src/main/java/org/nuxeo/ai/bulk/DataSetUploadComputation.java
covolution: You are right.  Its not needed so I will take it out.

nuxeo-ai-model/src/main/java/org/nuxeo/ai/bulk/DataSetUploadComputation.java
covolution: OK, just because you asked so nicely.

kevinleturc:
👍

kevinleturc:
👍

PR #37, opened 2019-01-03
NXP-26455: Add statistics to corpus dataset

covolution:


nuxeo-ai-model/src/main/resources/data/schemas/ai_corpus.xsd
```diff
@@ -39,17 +39,5 @@
...
-    </xs:sequence>
-  </xs:complexType>
-  <xs:complexType name="ai_corpus_features_histogramType">
-    <xs:sequence>
-      <xs:element name="field" type="xs:string"/>
-      <xs:element name="label" type="xs:string"/>
-      <xs:element name="count" type="xs:integer"/>
-    </xs:sequence>
-  </xs:complexType>
-  <xs:element name="features_histogram" type="nxs:ai_corpus_features_histogramListType"/>
```
PedroCardoso: Did you also remove the property on the studio ?

covolution: There is no studio project for the `nuxeo-ai` project.  For the studio project used by `nuxeo-ai-cloud` I just added the JSON blob.  I was wondering if we might post-process the json into a features histogram.

PR #41, opened 2019-02-01
NXP-26742: Add timeouts to CloudConfigDescriptor

covolution:


nuxeo-ai-model/src/main/java/org/nuxeo/ai/cloud/CloudConfigDescriptor.java
```diff
@@ -31,6 +31,12 @@
     @XNode("@url")
     protected String url = "http://localhost:8080/nuxeo";
 
+    @XNode("@readTimeout")
+    protected int readTimeout = 60 * 8; // Default to 8 minutes.
+
+    @XNode("@connectTimeout")
+    protected int connectTimeout = 15; // Default to 15 seconds.
```
efge: We can now use proper `Duration` contributions in descriptors, with a sane syntax for s/ms/min/etc. See 4876c41530c210cd95512bf90aec4160941384f5 for more. Please update your code to use it.

PR #45, opened 2019-02-13
NXP-26768: Improve stats query

covolution:


nuxeo-ai-model/src/main/java/org/nuxeo/ai/model/export/DatasetExportServiceImpl.java
```diff
@@ -246,6 +254,8 @@ protected Long getOverallStats(List<Map<String, String>> featuresWithType, List<
         for (Map<String, String> prop : featuresWithType) {
             String propName = prop.get(NAME_PROP);
             switch (prop.get(TYPE_PROP)) {
+                case CATEGORY_TYPE:
+                    // Drop through
```
kevinleturc: Don't you miss a `break` here ?

kevinleturc:
👍

nuxeo-ai-model/src/main/java/org/nuxeo/ai/model/export/DatasetExportServiceImpl.java
efge: Probably not given the `Drop through` comment, but I wouldn't event add the comment. This kind of comment is useful if there's some code then you all through the following case, but here it's just two cases:

                case CATEGORY_TYPE:
                case TEXT_TYPE:
                    code()

It's quite standard to just put the two cases one after the other.

covolution: I can change it.  With the comment I thought it was clearer, showing I was intentionally using the `case` logic below.

PR #69, opened 2019-05-21
NXP-27182: Add property ref content on suggestions

covolution:


covolution:
Partial output looks like this:
```  "property": "dc:subjects",
    "values": [{
        "name": "NO_MATCH",
        "confidence": "0.8650408",
        "asProperty": "NO_MATCH"
    }, {
        "name": "music",
        "confidence": "0.83437204",
        "asProperty": {
            "entity-type": "directoryEntry",
            "directoryName": "l10nsubjects",
            "id": "music",
            "properties": {
                "parent": "art",
                "ordering": 10000000,
                "obsolete": 0,
                "id": "music",
                "label_en": "Music",
                "label_fr": "Musique"
            }
        }
    }]
}]```

The `name` property could be considered a duplication.  I am currently keeping it for backward compatibility.

**Comment by Gabez0r** [comment]
As long as you don't use fetchers, `asProperty` should have the same value as `name`, so I think we do not need the duplication.

**Comment by covolution** [comment]
> As long as you don't use fetchers, `asProperty` should have the same value as `name`, so I think we do not need the duplication.

So the question is really: should we break backwards compatibility or not?

File: nuxeo-ai-model/src/main/java/org/nuxeo/ai/model/serving/SuggestionOp.java [hunk 1]
```diff
@@ -61,7 +83,82 @@ public Blob run(DocumentModel doc) throws IOException {
...
-        return Blobs.createJSONBlob(MAPPER.writeValueAsString(suggestions));
+
+        String toReturn = toJsonString(jg -> {
+            // JsonGenerator won't let you start with an array, so I am wrapping it in a field and removing the
+            // field at the end of the method.
+            jg.writeFieldName("return");
+            jg.writeStartArray();
+            suggestions.forEach(metadata -> {
+                try {
+                    DocumentModel input = coreSession.getDocument(new IdRef(metadata.context.documentRef));
```
andreinechaev: just for performance reasons I would consider `CoreSession#getDocuments` here, to basically pre-fetch. 
Although we know what there will be a small amount of suggestion, so not critical. Just a note

andreinechaev:
👍

PR #70, opened 2019-05-28
NXP-27437: Add an api and service to proxy calls to the cloud

covolution:
This api allows you to pass through get requests to project-specific models and datasets using the Nuxeo Java client.


andreinechaev:
👍

nuxeo-ai-jaxrs/src/test/resources/mappings/nuxeo_datasets.json
```diff
@@ -0,0 +1,14 @@
...
+    "headers": {
+      "Content-Type": "application/json"
+    }
+  }
+}
```
PedroCardoso: Probably nothing, but these seem special chars on the code . Maybe the compiler ignores, but maybe could cause problems !

PedroCardoso:
👍

PR #80, opened 2019-06-12
Feature nxp 27113 new dataset export admin page

andreinechaev:


covolution:
I don't understand this PR, lets discuss it.

covolution:
If we forced to use an operation then we should probably revert the changes I made for NXP-27437.

PR #82, opened 2019-06-20
NXP-27565: Report dataset export start/stop/errors 10.10

covolution:
This pull request needs to wait for the fix for NXP-27566 in 10.10-HF09 before merging, otherwise the tests will fail.


nuxeo-ai-core/src/test/java/org/nuxeo/ai/bulk/BadRecordWriter.java
```diff
@@ -32,8 +32,9 @@ public BadRecordWriter(String name, String bad) {
     }
 
     @Override
-    public void write(List<Record> list) {
+    public long write(List<Record> list) {
         // Will never get here
+        return 99;
```
andreinechaev: as a good practice should return a negative value

PR #86, opened 2019-07-12
Fix nxp 27413 autofill service

covolution:


nuxeo-ai-core/src/main/java/org/nuxeo/ai/auto/AutoService.java
```diff
@@ -0,0 +1,51 @@
...
+
+    /**
+     * Calculate all properties on a Document, given the type of calculation.
+     */
+    void calculateProperties(DocumentModel doc, AUTO_ACTION action);
+
+    /**
+     * Remove the property from the list of auto updated values and remove suggestions.
+     */
+    void confirmProperty(DocumentModel doc, String xpath);
```
PedroCardoso: Is this the best name ? I am terrible naming, but took me a while to get what it is.
How about updatedByHuman() or confirmedByHuman() confirmedByOperator()

nuxeo-ai-core/src/main/java/org/nuxeo/ai/auto/AutoService.java
```diff
@@ -0,0 +1,51 @@
...
+
+    /**
+     * Remove the property from the list of auto updated values.
+     */
+    void confirmProperty(DocumentModel doc, String xpath, boolean includeSuggestions);
```
PedroCardoso: same as above

nuxeo-ai-core/src/main/java/org/nuxeo/ai/auto/AutoServiceImpl.java
```diff
@@ -0,0 +1,168 @@
...
+    public void calculateProperties(DocumentModel doc) {
+        calculateProperties(doc, ALL);
+    }
+
+    @Override
+    public void calculateProperties(DocumentModel doc, AUTO_ACTION action) {
+        SuggestionMetadataAdapter adapted = doc.getAdapter(SuggestionMetadataAdapter.class);
+        for (String xpath : adapted.getSuggestedProperties()) {
+            switch (action) {
+                case ALL:
```
PedroCardoso: Aren't they exclusive ?
Auto-fill if no prior data is there, auto-correct if prior data is there.

nuxeo-ai-core/src/main/java/org/nuxeo/ai/auto/AutoService.java
covolution: I agree, its not actually implemented yet so I will change the name when we do the implementation.

nuxeo-ai-core/src/main/java/org/nuxeo/ai/auto/AutoService.java
covolution: I agree, its not actually implemented yet so I will change the name when we do the implementation.

nuxeo-ai-core/src/main/java/org/nuxeo/ai/auto/AutoServiceImpl.java
```diff
@@ -0,0 +1,168 @@
...
+        if (max != null) {
+            String comment = String.format("Auto filled %s. (Confidence %s , Threshold %s)",
+                                           xpath, max.getConfidence(), threshold);
+            log.debug(comment);
+            property.setValue(max.getName());
+            Framework.getService(DocMetadataService.class)
+                     .updateAuto(docMetadata.getDoc(), AUTO_FILLED, xpath, null, comment);
+        } else {
+            if (alreadyAutofilled) {
+                // We autofilled but now the value didn't autofill so lets reset it
```
PedroCardoso: Document in some place this behaviour, as this business logic is important to understand, and "accept".

nuxeo-ai-core/src/main/java/org/nuxeo/ai/auto/AutoServiceImpl.java
```diff
@@ -0,0 +1,168 @@
...
+                oldValue = null;
+            }
+            metadataService.updateAuto(docMetadata.getDoc(), AUTO_CORRECTED, xpath, oldValue, comment);
+            if (docMetadata.isAutoFilled(xpath)) {
+                // Autofill and Auto correct are mutually exclusive, we have auto corrected so lets remove autofill.
+                metadataService.resetAuto(docMetadata.getDoc(), AUTO_FILLED, xpath, false);
+            }
+        } else {
+            if (alreadyAutoCorrected) {
+                // We auto corrected but now the value didn't auto correct so lets reset it
```
PedroCardoso: Again, document this.

nuxeo-ai-core/src/main/java/org/nuxeo/ai/auto/AutoServiceImpl.java
covolution: Elsewhere in the code I mention auto-fill and auto-correct are mutually exclusive, but we still need to call them.  In the code, auto-correct is called before auto-fill - so autofill won't be calculated if autocorrect has set a value.

covolution: We need the possibliity to run both

nuxeo-ai-core/src/main/java/org/nuxeo/ai/auto/AutoServiceImpl.java
covolution: hmm, where's the best place for this?  I will add it to the readme

PR #92, opened 2019-07-24
Fix nxp 27554 suggestion removal

covolution:


nuxeo-ai-core/src/main/java/org/nuxeo/ai/services/DocMetadataServiceImpl.java
```diff
@@ -272,18 +269,100 @@ public DocumentModel resetAuto(DocumentModel doc, String autoField, String xPath
...
+    public static boolean hadBeenModified(DocumentModel doc, Set<String> props) {
+
+        if (props != null) {
+            for (String propName : props) {
+                try {
+                    Property prop = doc.getProperty(propName);
+                    if (prop != null && prop.isDirty()) {
+                        return true;
+                    }
+                } catch (PropertyNotFoundException e) {
```
andreinechaev: it's a runtime exception. is there a way to avoid that

covolution: Good question, its standard Nuxeo `PropertyNotFoundException`.  At some point we need to handle this. I am open to suggestions.

andreinechaev: idk. it's fine since the intention is very clear. Exceptions are not designed good in Nuxeo. Their usage contradicts with common practices.

PR #94, opened 2019-07-25
NXP-27488: extending ai corpus

andreinechaev:


nuxeo-ai-model/src/main/java/org/nuxeo/ai/adapters/AICorpus.java
```diff
@@ -0,0 +1,139 @@
...
+
+    public void setJobId(String id) {
+        doc.setPropertyValue(CORPUS_JOBID, id);
+    }
+
+    public String getJobId() {
+        return (String) doc.getPropertyValue(CORPUS_JOBID);
+    }
+
+    public static class IOParam extends HashMap<String, String> {
```
covolution: What's this?

nuxeo-ai-model/src/main/java/org/nuxeo/ai/adapters/Adapter.java
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * Contributors:
+ *     anechaev
+ */
+package org.nuxeo.ai.adapters;
+
+import org.nuxeo.ecm.core.api.CoreSession;
+import org.nuxeo.ecm.core.api.DocumentModel;
+
+public interface Adapter {
```
covolution: The name and concept of an `Adapter` is already used in the Nuxeo platform.  I don't understand why you created an interface with that name.

nuxeo-ai-model/src/main/java/org/nuxeo/ai/model/export/DatasetExportServiceImpl.java
```diff
@@ -120,22 +132,35 @@ public String export(CoreSession session, String nxql,
...
-        DocumentModel corpus = createCorpus(session, nxql, inputs, outputs, split, statsBlob);
+        AICorpus corpus = createCorpus(session, nxql, inputs, outputs, split, statsBlob);
 
         List<String> featuresList = new ArrayList<>(inputProperties);
         featuresList.addAll(outputProperties);
-        BulkCommand bulkCommand = new BulkCommand.Builder(EXPORT_ACTION_NAME, notNullNxql(nxql, featuresWithType))
+
+        String notNullNXQL = notNullNxql(nxql, featuresWithType);
+        String username = session.getPrincipal().getName();
+        BulkCommand bulkCommand = new BulkCommand.Builder(EXPORT_ACTION_NAME, notNullNXQL, username)
```
covolution: Need to check whether this constructor is available on 10.10

nuxeo-ai-model/src/main/java/org/nuxeo/ai/model/export/DatasetExportServiceImpl.java
```diff
@@ -293,4 +321,41 @@ protected String notNullNxql(String nxql, List<Map<String, String>> featuresWith
         }
         return buffy.toString();
     }
+
+    /**
+     * For the given property, find out if it exists and determine if its text or content
+     */
+    protected IOParam getPropertyWithType(String prop) {
```
covolution: Why did you move these methods?

nuxeo-ai-model/src/main/java/org/nuxeo/ai/adapters/AICorpus.java
andreinechaev: alias. this structure is used a lot including in other collections.

nuxeo-ai-model/src/main/java/org/nuxeo/ai/adapters/Adapter.java
andreinechaev: it's a default interface that allows to have consistency within the project. It requires derivatives to be able to give back document or other common props across the system.

nuxeo-ai-model/src/main/java/org/nuxeo/ai/model/export/DatasetExportServiceImpl.java
andreinechaev: it's not but the previous one was deprecated.

nuxeo-ai-model/src/main/java/org/nuxeo/ai/model/export/DatasetExportServiceImpl.java
andreinechaev: they were in another module but used in this service. I needed these helpers to be able to work with `IOParam` and to avoid cycling deps I moved them where they belong

nuxeo-ai-model/src/main/java/org/nuxeo/ai/adapters/AICorpus.java
andreinechaev: maybe better naming needed

nuxeo-ai-model/src/main/java/org/nuxeo/ai/model/export/DatasetExportServiceImpl.java
covolution: It seems like you are more likely to be creating a cyclic dependency.  The dependencies go : pipe -> core -> model -> package.  (Model was not included for 9.10 but after all, we didn't do a release for 9.10)

covolution: If you are saying its only used in 1 place and you are not duplicating code then I guess its ok to move

nuxeo-ai-model/src/main/java/org/nuxeo/ai/adapters/Adapter.java
covolution: Perhaps `AIAdapter` then?

nuxeo-ai-model/src/main/java/org/nuxeo/ai/model/export/DatasetExportServiceImpl.java
covolution: we must support 10.10

nuxeo-ai-model/src/main/java/org/nuxeo/ai/adapters/Adapter.java
andreinechaev: not really. It has nothing to do with AI more like a common interface that should be a common one within platform IMO

nuxeo-ai-model/src/main/java/org/nuxeo/ai/model/export/DatasetExportServiceImpl.java
andreinechaev: not arguing with that. It's master branch though. Aligned on non LTS Nuxeo and we use APIs that will be deprecated at some point. Just wanted to avoid that. Moving this part to 10.10 will requires some changes but at some point this difference will be there. So why not do it now?

covolution: AI is an addon so it doesn't necessaryily follow the versioning of the platform.  I am trying to keep our `master` branch compatible with 10.10

nuxeo-ai-model/src/main/java/org/nuxeo/ai/adapters/Adapter.java
covolution: That's my point.  Its already a concept in the platform but you are adding it in the AI module

andreinechaev: renamed to `AIAdapter`

nuxeo-ai-model/src/main/java/org/nuxeo/ai/model/export/DatasetExportServiceImpl.java
andreinechaev: moved back to deprecated constructor

PR #96, opened 2019-07-31
NXP-27705: Add a bulk enrichment action

covolution:


nuxeo-ai-model/src/main/java/org/nuxeo/ai/bulk/BulkEnrichmentAction.java
```diff
@@ -0,0 +1,109 @@
...
+        protected void compute(CoreSession coreSession, List<String> ids, Map<String, Serializable> map) {
+            ModelServingService modelServingService = Framework.getService(ModelServingService.class);
+            for (String id : ids) {
+                try {
+                    DocumentModel doc = coreSession.getDocument(new IdRef(id));
+                    Set<String> inputs = modelServingService.getInputs(doc);
+                    if (!inputs.isEmpty()) {
+                        BlobTextFromDocument blobTextFromDocument = PropertyUtils.docSerialize(doc, inputs);
+                        if (blobTextFromDocument != null) {
+                            outputs.add(toRecord(blobTextFromDocument.getKey(), blobTextFromDocument));
```
andreinechaev: how large outputs can be? Do we really need to keep the array on the class level?

nuxeo-ai-model/src/main/java/org/nuxeo/ai/bulk/DataSetBulkAction.java
```diff
@@ -132,7 +132,7 @@ protected void compute(CoreSession coreSession, List<String> ids, Map<String, Se
             for (String id : ids) {
                 try {
                     DocumentModel doc = coreSession.getDocument(new IdRef(id));
-                    BlobTextFromDocument subDoc = PropertyUtils.docSerialize(doc, customProperties);
+                    BlobTextFromDocument subDoc = PropertyUtils.docSerialize(doc, new HashSet<>(customProperties));
```
andreinechaev: why do you need to wrap it around? making it mutable?

nuxeo-ai-model/src/main/java/org/nuxeo/ai/bulk/BulkEnrichmentAction.java
covolution: Yes, that's a bit nasty isn't it.  Its the way other batch computations.  You need the `ComputationContext` to write the record.  The batch size is set by config, so its 50.

nuxeo-ai-model/src/main/java/org/nuxeo/ai/bulk/DataSetBulkAction.java
covolution: I changed the type from List to Set

andreinechaev:
👍

PR #99, opened 2019-08-06
NXP-27707: Bulk remove suggestions

covolution:


nuxeo-ai-model/src/main/java/org/nuxeo/ai/bulk/BulkRemoveEnrichmentAction.java
```diff
@@ -0,0 +1,151 @@
...
+            modelId = modelParam != null ? String.valueOf(modelParam) : null;
+            xPaths = xParam != null ? getSet(xParam) : Collections.emptySet();
+        }
+
+        @Override
+        protected void compute(CoreSession session, List<String> ids, Map<String, Serializable> properties) {
+            DocMetadataService metadataService = Framework.getService(DocMetadataService.class);
+            for (String id : ids) {
+                try {
+                    DocumentModel doc = session.getDocument(new IdRef(id));
```
andreinechaev: instead of capturing exception you can check it via `CoreSession#exists`

nuxeo-ai-core/src/main/java/org/nuxeo/ai/auto/AutoServiceImpl.java
```diff
@@ -157,11 +163,14 @@ protected void calculateAutoCorrect(SuggestionMetadataWrapper docMetadata, Strin
...
     }
 
-    protected boolean setProperty(Property property, String newValue) {
-        Serializable currentValue = property.getValue();
+    protected boolean setProperty(CoreSession session, DocumentModel doc, String key, Serializable currentValue, String newValue) {
         if (!newValue.equals(currentValue)) {
-            property.setValue(newValue);
-            return true;
+            try {
+                DocumentHelper.setProperty(session, doc, key, newValue);
```
andreinechaev: a bit confused why do we need this helper when setting a property is very straightforward and the advantage of such is questionable

nuxeo-ai-core/src/main/java/org/nuxeo/ai/functions/SaveEnrichmentFunction.java
```diff
@@ -41,7 +42,11 @@ public void accept(EnrichmentMetadata metadata) {
                     DocMetadataService docMetadataService = Framework.getService(DocMetadataService.class);
                     DocumentModel doc = docMetadataService.saveEnrichment(session, metadata);
                     if (doc != null) {
-                        session.saveDocument(doc);
+                        try {
+                            session.saveDocument(doc);
```
andreinechaev: why not rethrowing the exception? am I missing something, why swallow?

nuxeo-ai-model/src/main/java/org/nuxeo/ai/bulk/BulkRemoveEnrichmentAction.java
```diff
@@ -0,0 +1,151 @@
...
+                            removalProperties = suggestions.stream()
+                                                           .map(LabelSuggestion::getProperty)
+                                                           .collect(Collectors.toSet());
+                        } else if (!xPaths.isEmpty()) {
+                            removalProperties = xPaths;
+                        } else {
+                            removalProperties = wrapper.getAutoProperties();
+                        }
+
+                        for (String xPath : removalProperties) {
```
andreinechaev: nested loop is the only one thing that bothers me here.

nuxeo-ai-model/src/main/java/org/nuxeo/ai/bulk/BulkRemoveEnrichmentAction.java
covolution: Would that then be less efficient because its making 2 calls?

nuxeo-ai-core/src/main/java/org/nuxeo/ai/auto/AutoServiceImpl.java
covolution: I am using it because it supports multi-value fields.

nuxeo-ai-core/src/main/java/org/nuxeo/ai/functions/SaveEnrichmentFunction.java
covolution: Each call is atomic, I came across this error now that autofill properties are being generated.  If the error is thrown up the stack then generic streams re-try functionality kicks in, which is not what we want.

nuxeo-ai-model/src/main/java/org/nuxeo/ai/bulk/BulkRemoveEnrichmentAction.java
andreinechaev: it's an interesting question. 
- Don't think so, first retrieving the doc should be O(1). second I'd assume that just the creation of exception is a relatively expensive process and it depends on how deep the stack trace is.

nuxeo-ai-core/src/main/java/org/nuxeo/ai/auto/AutoServiceImpl.java
andreinechaev: ok

nuxeo-ai-core/src/main/java/org/nuxeo/ai/functions/SaveEnrichmentFunction.java
andreinechaev: ok

nuxeo-ai-model/src/main/java/org/nuxeo/ai/bulk/BulkRemoveEnrichmentAction.java
covolution: Found the `loadDocuments` method instead

nuxeo-ai-model/src/main/java/org/nuxeo/ai/bulk/BulkRemoveEnrichmentAction.java
covolution: The inner loop is just a small Set of property names, probably less than 5 values.

PR #100, opened 2019-08-06
NXP-27225: merge dirty properties received on suggestion op

nmpcunha:
Disclaimer:
- Not sure if I have the correct formatter or not, but some of the code got reformatted. Please let me know if which formatter should I use;
- The second commit was an attempt to test the properties merge, which lead me to the need of using Mocks and ArgumentCaptors. In theory, by using that kind of approach the test will be unitary and not a kind of integration test. Still, please let me know if you have other approach for ensuring that the new code is properly tested;

Thanks!

### DIFF
--- a/nuxeo-ai-model/src/main/java/org/nuxeo/ai/model/serving/SuggestionOp.java
+++ b/nuxeo-ai-model/src/main/java/org/nuxeo/ai/model/serving/SuggestionOp.java
@@ -22,15 +22,18 @@ import static org.nuxeo.ai.pipes.services.JacksonUtil.MAPPER;
 
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
+import java.io.Serializable;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Objects;
 import java.util.stream.Collectors;
+
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.nuxeo.ai.enrichment.EnrichmentMetadata;
 import org.nuxeo.ai.metadata.AIMetadata;
 import org.nuxeo.ai.metadata.LabelSuggestion;
-import org.nuxeo.ai.enrichment.EnrichmentMetadata;
 import org.nuxeo.ecm.automation.core.Constants;
 import org.nuxeo.ecm.automation.core.annotations.Context;
 import org.nuxeo.ecm.automation.core.annotations.Operation;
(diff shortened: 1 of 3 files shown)


nuxeo-ai-model/src/main/java/org/nuxeo/ai/model/serving/SuggestionOp.java
```diff
@@ -77,27 +80,40 @@
...
+    @OperationMethod
+    public Blob run(DocumentRef docRef) {
+        DocumentModel docModel = coreSession.getDocument(docRef);
+        return run(docModel);
+    }
+
     @OperationMethod
     public Blob run(DocumentModel doc) {
+        if (updatedDocument != null) {
+            Arrays.asList(doc.getSchemas())
```
andreinechaev: please no nested loops:

```
Arrays.stream(doc.getSchemas())
                            .flatMap(s1 -> udoc.getProperties(s1).entrySet().stream())
                            .forEach(entry -> doc.setPropertyValue(entry.getKey(), (Serializable) entry.getValue()));
```

should do the job

nuxeo-ai-model/src/main/java/org/nuxeo/ai/model/serving/SuggestionOp.java
```diff
@@ -77,27 +80,40 @@
...
+                  .forEach(schema -> updatedDocument.getProperties(
+                          schema).forEach((key, value) -> doc.setPropertyValue(key, (Serializable) value)));
+        }
 
-        List<EnrichmentMetadata> suggestions;
-        if (doc == null || (suggestions = modelServingService.predict(doc)) == null || suggestions.isEmpty()) {
```
covolution: Not sure why you changed this?  The current style was suggested in a code review

nuxeo-ai-model/src/main/java/org/nuxeo/ai/model/serving/SuggestionOp.java
covolution: Personally, in this example, I think a standard for loop will be much clearer

andreinechaev: depends, if nested loops don't scare you, then yes :)

nuxeo-ai-model/src/main/java/org/nuxeo/ai/model/serving/SuggestionOp.java
andreinechaev: cleaner, I'd assume

nuxeo-ai-model/src/main/java/org/nuxeo/ai/model/serving/SuggestionOp.java
nmpcunha: Thanks for your feedback!
Not afraid by nested loops (only two in fact), but I can understand your point @andreinechaev about the usage of streams here. I'll perform the change

nuxeo-ai-model/src/main/java/org/nuxeo/ai/model/serving/SuggestionOp.java
nmpcunha: Two reasons:
- Initially for my mind map and for debugging;
- Then I found cleaner, more readable;

But I can change it to the original one.

nmpcunha:
Regarding your question about real end to end test, the answer is no.
In theory, unit tests shouldn't rely on real services. In this specific case the change will update some properties on the document that is sent to the prediction service. So, that should be what we test, no? And that's why I created other class in order to have something more concrete to show and discuss.
How would you suggest to test this specific change, @covolution ?

nuxeo-ai-model/src/main/java/org/nuxeo/ai/model/serving/SuggestionOp.java
covolution: Leave it, how you did it.

PR #101, opened 2019-08-07
NXP-27225: merge dirty properties received on suggestion op (Backport 10.10)

nmpcunha:


nmpcunha:
I did the needed changes for backporting to 10.10. If we are thinking to change the syntax, should those changes land on your PR?
Or are you trying to have 10.10 and master in sync in commits? Because having web-ui files in this repo won't allow it, since we'll always have more commits in this branch.

PR #103, opened 2019-08-12
NXP-27489: tracking export start/end time

andreinechaev:


nuxeo-ai-model/src/main/java/org/nuxeo/ai/cloud/AICorpus.java
```diff
@@ -0,0 +1,227 @@
...
+        protected long evaluationDocCount;
+
+        @JsonProperty("query")
+        protected String query;
+
+        @JsonProperty("split")
+        protected int split;
+
+        @JsonProperty("fields")
+        protected List<Map<String, Object>> fields;
```
PedroCardoso: Why `Object` as a type in `Map<String, Object>`
The filed should be a field name and type, being type a enum or a static String.

nuxeo-ai-model/src/main/java/org/nuxeo/ai/cloud/AICorpus.java
```diff
@@ -0,0 +1,227 @@
...
+ */
+package org.nuxeo.ai.cloud;
+
+import java.util.Date;
+import java.util.List;
+import java.util.Map;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class AICorpus {
```
covolution: You just renamed `AICorpus` to `DatasetExport`, can we give this a better name and some javadoc?

covolution: e.g. POJO for a DocumentType in the cloud

andreinechaev: done

andreinechaev: it's AI_Corpus, I'm not sure what can be a better name :)

PR #105, opened 2019-08-19
NXP-27784: Don't export auto properties

covolution:


nuxeo-ai-model/src/main/java/org/nuxeo/ai/bulk/DataSetBulkAction.java
```diff
@@ -129,26 +130,31 @@ protected void compute(CoreSession coreSession, List<String> ids, Map<String, Se
...
-                        Record record = toRecord(command.getId(), subDoc);
-                        if (isTraining) {
-                            training.add(record);
-                        } else {
-                            validation.add(record);
-                        }
+            for (DocumentModel doc : loadDocuments(coreSession, ids)) {
+                BlobTextFromDocument subDoc = null;
+                List<String> targetProperties = new ArrayList<>();
+                targetProperties.addAll(customProperties);
```
andreinechaev: `ArrayList` constructor accepts another collection to create a copy from it.

covolution: Yes, I started with that, but had a problem.  One of them was unmodifiable so couldn't copy - it was strange.

andreinechaev: hm, ok. my bad

covolution: actually, i didn't understand it but changed the code to be sure.

PR #108, opened 2019-08-28
NXP-27880: contribute aiModels enricher to WebUI config

nmpcunha:


nuxeo-ai-core/src/main/resources/META-INF/MANIFEST.MF
```diff
@@ -11,4 +11,5 @@ Nuxeo-Component: OSGI-INF/ai-contrib.xml,
  OSGI-INF/ai-metadata-service.xml,
  OSGI-INF/ai-auto-contrib.xml,
  OSGI-INF/ai-auto-threshold.xml,
+ OSGI-INF/ai-web-ui-enrichment.xml,
```
Gabez0r: Maybe?
```suggestion
 OSGI-INF/ai-web-ui-properties.xml,
```
It would be consistent with the naming we use on Web UI.

nmpcunha: 👍

PR #123, opened 2019-11-04
NXP-28184: video captions facet

andreinechaev:


nuxeo-ai-core/src/main/resources/OSGI-INF/caption-core-type-contrib.xml
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0"?>
+<component name="org.nuxeo.ecm.ai.caption.schema">
+
+  <require>org.nuxeo.ecm.core.CoreExtensions</require>
```
vpasquier: I think (not sure) we need to require the def of Video first: `org.nuxeo.nuxeo.ecm.platform.video.doctype` to be able to "append"

PedroCardoso:
The Issue number is not correct. Trying to find reference to the documentation with the Facet properties.

PR #134, opened 2019-12-05
NXP-28392: don't check schema on SimpleDoc

andreinechaev:


andreinechaev:
because the updated doc coming from UI and treated as a SimpleDocument which is not aware of schemas. Adding it I didn't know this fact

PR #136, opened 2019-12-10
NXP-28404 template for batch processing

andreinechaev:


nuxeo-ai-core-package/src/main/resources/install/templates/nuxeo-ai-core/nuxeo.defaults
```diff
@@ -14,3 +14,8 @@ nuxeo.enrichment.save.facets=true
 nuxeo.enrichment.raiseEvent=true
 
 nuxeo.ai.export.tensorflow=true
+
+# Configuration for Export Stream
```
vpasquier: We should set values by default

vpasquier: 100 for validation and 200 for training - 50 for batches?

PR #142, opened 2020-01-02
NXP-28479: cleanup logs - rewrite dataset split

vpasquier:


nuxeo-ai-model/src/main/java/org/nuxeo/ai/bulk/DataSetBulkAction.java
```diff
@@ -127,41 +128,56 @@ public ExportingComputation(String name) {
...
-                    targetProperties.removeAll(wrapper.getAutoCorrected());
-                }
 
-                BlobTextFromDocument subDoc = null;
-                if (!targetProperties.isEmpty()) {
-                    subDoc = PropertyUtils.docSerialize(doc, new HashSet<>(targetProperties));
-                }
+            // Split documents list into training and validation datasets
+            DocumentModelList docs = loadDocuments(coreSession, ids);
+            List<List<DocumentModel>> datasets = ListUtils.partition(docs, (docs.size() - 1) * percentSplit / 100);
```
PedroCardoso: This will not work. We need a uniform sampling over time. If we split 20 from the beginning or end could bring evaluation errors later

nuxeo-ai-core/src/main/java/org/nuxeo/ai/bulk/AbstractRecordWriter.java
```diff
@@ -136,9 +136,6 @@ protected File getFile(String id) {
                 throw new NuxeoException("Unable to create a temp file. ", e);
             }
         } else {
-            if (log.isDebugEnabled()) {
```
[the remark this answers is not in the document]
vpasquier: I didn't find this useful + it's a real pollution (350k logs like this just for revlon on two days instance up)

nuxeo-ai-model/src/main/java/org/nuxeo/ai/bulk/DataSetBulkAction.java
vpasquier: How could it work before so?

nuxeo-ai-core/src/main/java/org/nuxeo/ai/bulk/AbstractRecordWriter.java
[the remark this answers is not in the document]
PedroCardoso: ok

nuxeo-ai-model/src/main/java/org/nuxeo/ai/bulk/DataSetBulkAction.java
PedroCardoso: As talked, shuffle all before a data cut

PR #157, opened 2020-02-25
NXP-28698: fixing client reinit

andreinechaev:


nuxeo-ai-model/src/main/java/org/nuxeo/ai/cloud/NuxeoCloudClient.java
```diff
@@ -98,10 +98,12 @@ public void start(ComponentContext context) {
...
-        if (configs.size() == 1) {
-            configureClient(configs.get(0));
-        } else {
-            throw new IllegalArgumentException("Nuxeo cloud client requires 1 single configuration.");
+        if (!configs.isEmpty()) {
+            if (configs.size() == 1) {
+                configureClient(configs.get(0));
+            } else {
+                throw new IllegalArgumentException("Nuxeo cloud client requires 1 single configuration.");
+            }
```
jcarsique: you can fully inline it

```java
if (configs.isEmpty()) {
    // NOP
} else if (configs.size() == 1) {
    configureClient(configs.get(0));
} else {
    throw new IllegalArgumentException("Nuxeo cloud client requires 1 single configuration.");
}

andreinechaev: it's more like it was initially done by Gethin. No real preferences here on my side

PR #164, opened 2020-03-02
NXP-28453: AI optimized image resizer

andreinechaev:


vpasquier:
I approved the resizer (but i guess it has to be changed to be used for the AI custom rendition)

nuxeo-ai-core/src/main/java/org/nuxeo/ai/convert/AiImageResizer.java
```diff
@@ -0,0 +1,107 @@
...
+ *     anechaev
+ *
+ */
+package org.nuxeo.ai.convert;
+
+import static org.nuxeo.ai.convert.AiResizePictureConverter.AI_JPEG_RESIZER_COMMAND;
+import static org.nuxeo.ai.convert.AiResizePictureConverter.AI_RESIZER_COMMAND;
+import static org.nuxeo.ecm.platform.picture.api.ImagingConvertConstants.JPEG_CONVERSATION_FORMAT;
+
+import java.awt.*;
```
PedroCardoso: I believe we do not want * as import

andreinechaev: correct

PR #181, opened 2020-04-17
AICORE-285: supersede es backport

vpasquier:


andreinechaev:
Just out of curiosity. Didn't you need to change some tests so they can work with ES page provider?

vpasquier:
> Just out of curiosity. Didn't you need to change some tests so they can work with ES page provider?

No we are not using this pp in the project itself (only on insight UI side)

PR #183, opened 2020-04-20
AICORE-226: add link in new admin page of aicore

vpasquier:


nuxeo-ai-model/src/main/java/org/nuxeo/ai/model/serving/FetchInsightURI.java
```diff
@@ -0,0 +1,49 @@
...
+public class FetchInsightURI {
+
+    public static final String ID = "AI.FetchInsightURI";
+
+    @OperationMethod
+    public String run() {
+        CloudClient cloudClient = Framework.getService(CloudClient.class);
+        CloudConfigDescriptor cloudConfig = cloudClient.getCloudConfig();
+        if (cloudConfig == null) {
+            throw new DocumentNotFoundException("Client Cloud configuration is missing");
```
andreinechaev, line 42: Are you sure that `DocumentNotFoundException` is the best here? Why not simple `OperationException` with your definitive message

nuxeo-ai-model/src/main/java/org/nuxeo/ai/model/serving/FetchInsightURI.java
```diff
@@ -0,0 +1,49 @@
...
+        }
+        String projectId = cloudConfig.getProjectId();
+        String url = cloudConfig.getUrl();
+        String token = cloudConfig.getAuthentication().getToken();
+        return String.format("/ai%s?token=%s&projectid=%s", url, token, projectId);
```
andreinechaev, line 47: I would imagine it would be better to return an object with such parameters and than the requester is responsible to construct such a request. What do you think?

nuxeo-ai-model/src/main/java/org/nuxeo/ai/model/serving/FetchInsightURI.java
vpasquier: Cause i want to use the 404 not found code return.... Couldn't find a way to customize it. (some nuxeo documentation are not up to date)

nuxeo-ai-model/src/main/java/org/nuxeo/ai/model/serving/FetchInsightURI.java
vpasquier: I was doing that and told to myself that for 3 values, we could not care much. Wdyt?

nuxeo-ai-web-ui/src/main/resources/web/nuxeo.war/ui/nuxeo-ai/admin/nuxeo-admin-insight.html
```diff
@@ -0,0 +1,70 @@
...
+    Polymer({
+      is: 'nuxeo-admin-insight',
+      behaviors: [Nuxeo.I18nBehavior],
+      properties: {
+        link: {
+          type: String,
+          computed: '_getInsightURL()',
+        },
+      },
+      _getInsightURL: function () {
```
nelsonsilva, line 61:
```suggestion
      ready: function () {
        this.$.aiInsight.execute().then(({ url, token, projectId, urlCore}) => this.link = `${url}?token=${token}&projectid=${projectId}&urlCore=${urlCore}`);
      }
```

PR #192, opened 2020-05-20
AICORE-287: obtaining field statistics for UI

andreinechaev:


nuxeo-ai-model/src/main/java/org/nuxeo/ai/model/analyzis/DatasetStatsService.java
```diff
@@ -0,0 +1,119 @@
...
+        if (field != null && isVocabulary(field.getType())) {
+            type = CATEGORY_TYPE;
+        } else if (field != null && TypeConstants.isContentType(field.getType())) {
+            type = IMAGE_TYPE;
+        } else {
+            type = TEXT_TYPE;
+        }
+        return type;
+    }
+}
```
PedroCardoso, line 119: This is the stats part that will need to improve. assuming text or image based on no real info.
But, I guess that to general stats, all that we need to know is if it is blob or not

andreinechaev: it is a field definition that can be changed during model creation.

PR #196, opened 2020-06-01
AICORE-302: reverting aggregates for text data types

andreinechaev:


nuxeo-ai-core/src/test/java/org/nuxeo/ai/rest/TestRestEnrichment.java
```diff
@@ -45,13 +48,19 @@
...
 
 @RunWith(FeaturesRunner.class)
-@Features({EnrichmentTestFeature.class, PlatformFeature.class})
-@Deploy({"org.nuxeo.ai.ai-core:OSGI-INF/enrichment-rest-test.xml"})
+@Features({ EnrichmentTestFeature.class, PlatformFeature.class })
+@Deploy({ "org.nuxeo.ai.ai-core:OSGI-INF/enrichment-rest-test.xml" })
 public class TestRestEnrichment {
 
+    @Rule
+    public WireMockRule wmr = new WireMockRule(options().extensions(new ResponseTemplateTransformer(true)).port(5089));
```
andreinechaev, line 62: `explorer.nuxeo.com` started to return `500`, therefore breaking tests. It shouldn't be used in first place but history remains history

nuxeo-ai-model/src/main/java/org/nuxeo/ai/model/export/DatasetExportServiceImpl.java
```diff
@@ -262,9 +262,10 @@ protected void validateParams(String nxql, Collection<String> inputProperties,
     protected void getValidStats(List<PropertyType> featuresWithType, List<Statistic> stats, NxQueryBuilder qb) {
         for (PropertyType prop : featuresWithType) {
             String propName = prop.getName();
-            if (IMAGE_TYPE.equals(prop.getType()) || TEXT_TYPE.equals(prop.getType())) {
+            if (IMAGE_TYPE.equals(prop.getType())) {
                 // qb.addAggregate(makeAggregate(AGG_CARDINALITY, contentProperty(propName), EMPTY_PROPS));
             } else {
+                // Apply aggregates for Text and Category data types
```
PedroCardoso, line 268: I wonder if it is not better really define that it is for Text and Categories, in case we have new types in the future 
(and we might have numeric classes)

andreinechaev: yep, sure

PR #203, opened 2020-06-15
AICORE-311: skip query modification for everything but images

andreinechaev:


nuxeo-ai-model/src/main/java/org/nuxeo/ai/model/export/DatasetExportServiceImpl.java
```diff
@@ -324,10 +324,10 @@ protected String notNullNxql(String nxql, List<PropertyType> featuresWithType) {
             String propName = prop.getName();
             if (IMAGE_TYPE.equals(prop.getType())) {
                 sb.append(" AND ").append(contentProperty(propName)).append(" IS NOT NULL");
-            } else if (CATEGORY_TYPE.equals(prop.getType())) {
+            } else if (CATEGORY_TYPE.equals(prop.getType()) || TEXT_TYPE.equals(prop.getType())) {
```
PedroCardoso, line 327: This will be a problem When the type is text, and the property is a Blob to be converted. 
You need to find out what kind of property it is.

nuxeo-ai-model/src/test/java/org/nuxeo/ai/model/TestDatasetOperation.java
```diff
@@ -146,8 +147,16 @@ public void shouldRunRestartOperation() throws OperationException {
...
+
+        Properties inputs = new Properties();
+        inputs.put("dc:title", "txt");
+        inputs.put("dc:description", "txt");
+        params.put("inputProperties", inputs);
+
+        Properties outputs = new Properties();
+        outputs.put("dc:nature", "cat");
+        params.put("outputProperties", outputs);
+
```
PedroCardoso, line 159: Have a property that will be a blob.

nuxeo-ai-model/src/main/java/org/nuxeo/ai/model/export/DatasetExportServiceImpl.java
andreinechaev: Wizard allows to set any type to any property

PedroCardoso: I know, that is why you need to check here if it is a string, or a blob.
In case of a blob, handle the same way as an image type

PedroCardoso: There is an util function to check if it is a blob property

nuxeo-ai-model/src/test/java/org/nuxeo/ai/model/TestDatasetOperation.java
andreinechaev: blob processing covered in a different place, it's simple rewrite of deprecated properties

PR #210, opened 2020-06-23
AICORE-317: simplifying statistics to get terms and cardinality

andreinechaev:


nuxeo-ai-model/src/main/java/org/nuxeo/ai/model/export/DatasetExportServiceImpl.java
```diff
@@ -315,22 +300,25 @@ protected Long getOverallStats(List<PropertyType> featuresWithType, List<Statist
                 case TEXT_TYPE:
                     // TODO assuming that text is a property ! could be a blob
                     qb.addAggregate(makeAggregate(AGG_MISSING, propName, EMPTY_PROPS));
+                    qb.addAggregate(makeAggregate(AGG_TYPE_TERMS, propName, TERM_PROPS));
```
PedroCardoso, line 303: We do not need terms stats for Text fields.

andreinechaev: actually we do. If you take a look at where I removed them, we had it before. And yes - it's a UI case. But it's still a case

PR #231, opened 2020-07-15
AICORE-336: add log4j loggers in package

vpasquier:


jcarsique:
See `<appender namespace="org.nuxeo.ai.cloud" severity="TRACE"/>` 
versus `<Logger name="org.nuxeo.ai.cloud" level="trace">`

The naming is not correct:
- the appenders (logs output) are likely already configured and we don't need/want to configure them (that could be source of debate around security).
- here, you are configuring a `logger`, not an  ̀appender`
- `namespace` is a `name` (formerly a `category` in Log4J 1)
- `severity` is the `level`

Ideally, I would look at contributing a native Log4J configuration file/content (usual XML, new YAML or old-style properties) to be merged with the runtime configuration.
That would avoid to implement a wrapper and would be much more flexible.

But contributing only name/category and level is enough to match the need :+1: 

You can look at nuxeoctl code for dynamic contribution to Log4J, although this was written for v1 and we upgraded to v2.

vpasquier:
Change has been made - i will add certainly the appender target maybe (if it will be clear enough)

PR #233, opened 2020-07-15
AICORE-329: add preview for ai core

vpasquier:


charts/preview/values.yaml
```diff
@@ -0,0 +1,62 @@
...
+        fabric8.io/ingress.name: preview
+        fabric8.io/ingress.annotations: |-
+          nginx.ingress.kubernetes.io/proxy-body-size: 0
+          nginx.ingress.kubernetes.io/server-snippet: ignore_invalid_headers off;
+      labels:
+        branch: "$BRANCH_NAME"
+        team: ai
+        resource: service
+        usage: preview
+    packages: nuxeo-dam nuxeo-ai-core
```
jcarsique, line 40: don't we need to set the wanted package versions here?

jcarsique:
Continued in #234 (branch renamed)

PR #257, opened 2020-10-12
AICORE-355 and AICORE-363

vpasquier:


nuxeo-ai-core/src/main/java/org/nuxeo/ai/services/PersistedConfigurationServiceImpl.java
```diff
@@ -0,0 +1,116 @@
...
+    private static final Logger log = LogManager.getLogger(PersistedConfigurationServiceImpl.class);
+
+    private static final ExtensionDescriptorReader reader = new ExtensionDescriptorReader();
+
+    public static final String KEY_VALUE_STORE = "aiConfigurationKVStore";
+
+    public static final String ALL_KEYS = "KEYS";
+
+    public PersistedConfigurationServiceImpl() {
+        this.register(ThresholdConfiguratorDescriptor.class);
```
vpasquier, line 54: this part will have to be more generic - i let it this way for now (except if better idea ?)

nuxeo-ai-core/src/main/java/org/nuxeo/ai/services/PersistedConfigurationServiceImpl.java
```diff
@@ -0,0 +1,116 @@
...
+
+    @Override
+    public Pair<String, List<Descriptor>> retrieveAllDescriptors() throws IOException {
+        Set<String> keys = getAllKeys();
+        List<Descriptor> descriptors = new ArrayList<>();
+        String xmlPayLoad = "";
+        for (String key : keys) {
+            byte[] bytes = getStore().get(key);
+            XMap xmap = reader.getXMap();
+            xmlPayLoad.concat(new String(bytes));
```
vpasquier, line 99: I guess this needs to be close by xml anchor... Some tests are in progress to make that sure

jcarsique:
Force-pushed after rebase on sprint-3a-10.10

nuxeo-ai-core/src/test/java/org/nuxeo/ai/configuration/PersistedConfigurationServiceTest.java
```diff
@@ -115,13 +120,31 @@ public void shouldGetAllConfigsByClass() throws IOException {
...
-        messageReceivedLatch = new CountDownLatch(1);
-        aiConfigurationService.set(threshold);
-        if (!messageReceivedLatch.await(5, TimeUnit.SECONDS)) {
-            fail("message not received in 5s");
+        int thresholdSize = ((ThresholdComponent) Framework.getRuntime()
+                                                           .getComponent(
+                                                                   "org.nuxeo.ai.configuration.ThresholdComponent")).typeThresholds.size();
+        // TODO: AICORE-366
+        // messageReceivedLatch = new CountDownLatch(1);
+        aiConfigurationService.set(thresholdFolder);
```
andreinechaev, line 128: It's not enough for our tests?

addons/nuxeo-ai-aws-core/src/main/java/org/nuxeo/ai/endpoint/Rekognition.java
```diff
@@ -60,8 +60,8 @@
 /**
  * Endpoints responsible for listening notifications from external services such as AWS SNS
  */
-@Path("/ai/rekognition")
-@WebObject(type = "ai")
+@Path("/aiaddons/rekognition")
```
andreinechaev, line 63: I think we should try to make the API common throughout the add-on. We don't need to cross import the projects. As you remember we can spawn a web object by its type, like we do in the [cloud](https://github.com/nuxeo/nuxeo-ai-cloud/blob/master/ai-cloud-jaxrs/src/main/java/org/nuxeo/ecm/restapi/server/jaxrs/AIWebObject.java#L430). 
Then we'd have endpoints `ai/config/...` and `ai/rekognition`, and so on

vpasquier: I tried and like i told you we cant put a dependency on nuxeo-ai-config

nuxeo-ai-core/src/test/java/org/nuxeo/ai/configuration/PersistedConfigurationServiceTest.java
vpasquier: I would prefer in general to have a lock rather than timeout

addons/nuxeo-ai-aws-core/src/main/java/org/nuxeo/ai/endpoint/Rekognition.java
andreinechaev: but we don't need to. WebObject will be resolved by simple string name

vpasquier: You need to return the web adapter of aws in the AIRoot.

PR #263, opened 2020-10-16
AICORE-365: calling suggestion op on aiModeled docs

andreinechaev:


vpasquier (approved):

nuxeo-ai-web-ui/src/main/resources/web/nuxeo.war/ui/nuxeo-ai/nuxeo-ai.html
```diff
@@ -175,7 +175,7 @@
         if (widget && Array.isArray(widget.suggestions) && widget.suggestions.length > 0) {
           widget._matchInput();
         }
-        if (noPropertyChanged || isModelInput) {
+        if (noPropertyChanged && isModelInput) {
```
nmpcunha, line 178: Not sure if I'm understanding the change, TBH. (guess I'm lacking some context).
Changing to `&&` will probably turn the `noPropertyChanged` obsolete, right? (or at least part of it, since `path` && `!path` are opposite)

nmpcunha (commented):
Not sure if I can help much more without context, but the change itself seems to be "hiding" something else that we may be missing 🤔

nuxeo-ai-web-ui/src/main/resources/web/nuxeo.war/ui/nuxeo-ai/nuxeo-ai.html
andreinechaev: suggestion operation called in this block should be called only when there is `aiModel` in the context of the doc. Also, as I understand this `noPropertyChange` was made to ensure that we call it only on clean Documents. In terms of that it's called only when Document Create/Edit view is just opened. Or am I wrong? I feel a bit confused that's why I summoned you here :)

PR #265, opened 2020-10-20
AICORE-367: add new endpoints and services for Model Definitions

vpasquier:


nuxeo-ai-config/pom.xml
```diff
@@ -25,10 +29,6 @@
       <groupId>org.nuxeo.ecm.automation</groupId>
       <artifactId>nuxeo-automation-core</artifactId>
     </dependency>
-    <dependency>
-      <groupId>org.nuxeo.ai</groupId>
-      <artifactId>nuxeo-ai-core</artifactId>
-    </dependency>
```
jcarsique, line 31: It's needed for a dependency on EnrichmentTestFeature

jcarsique (changes requested):

nuxeo-ai-config/pom.xml
vpasquier: i removed it cause i duplicated it previously in the pom

vpasquier: But you're right actually cause i didnt set the test-jar hehe. Thanks!

jcarsique (approved):

PedroCardoso (approved):

PedroCardoso:
Would be nice to have a mock service to call insight with a mock model. And test calls after adding the model xml .

PR #268, opened 2020-10-22
fix AICORE-376 pkgVersion

jcarsique:
Enforce fixed downstream package version to help circumvent NXP-29794


tiagocardoso (approved):
Looks like it needed the version var to be with that syntax (double). 
I don't know, but I thrust you ;)

jcarsique:
> Looks like it needed the version var to be with that syntax (double).

@tiagocardoso the pattern is `<package name>:<min version>:<max version>`
While it's recommended to only set the min version in order to allow minor upgrades, the above strict requirement will partly prevent NXP-29794 from introducing an unwanted greater snapshot version (that should normally be excluded).

PR #272, opened 2020-10-26
AICORE-380: add continuous export management apis

vpasquier:


jcarsique (approved):

nuxeo-ai-core-package/src/main/resources/install/templates/nuxeo-ai-core/config/ai-config.xml.nxftl
```diff
@@ -45,4 +45,14 @@
   </extension>
 </#if>
 
+<#if "${nuxeo.ai.continuous.export.enable}" == "true">
+  <extension target="org.nuxeo.ecm.core.scheduler.SchedulerService" point="schedule">
+    <schedule id="continuous_export_default">
+        <event>startContinuousExport</event>
+        <!-- At 03:00 AM, every 7 days -->
```
andreinechaev, line 52: please remove the comment

andreinechaev (approved):

PedroCardoso (approved):

nuxeo-ai-config/src/test/java/org/nuxeo/ai/TestAIConfigREST.java
```diff
@@ -104,15 +106,16 @@ public void iCanSetNuxeoConfVar() throws IOException {
...
             assertThat(response.getStatus()).isEqualTo(Response.Status.OK.getStatusCode());
             assertThat(Framework.getProperty("test")).isNotNull().isNotEmpty();
         }
     }
 
     @Test
     public void iCanSetThreshold() {
-        try (CloseableClientResponse response = getResponse(RequestType.POST, "ai/extension/thresholds", thresholdFile,
+        try (CloseableClientResponse response = getResponse(RequestType.POST,
+                "aicore/extension/thresholds?docType=File", thresholdFile,
```
jcarsique, line 118: The test fails https://jenkins.ai.dev.nuxeo.com/blue/organizations/jenkins/nuxeo%2Fnuxeo-ai/detail/PR-272/5/tests

nuxeo-ai-core/src/main/java/org/nuxeo/ai/services/PersistedConfigurationServiceImpl.java
```diff
@@ -104,22 +124,41 @@ public Descriptor retrieve(String key) throws IOException {
...
-            builder.append(xml);
+        DocumentBuilder docBuilder;
+        Element root;
+        try {
+            docBuilder = dbfac.newDocumentBuilder();
+            Document doc = docBuilder.newDocument();
+            root = doc.createElement(tag);
+            for (Descriptor descriptor : descriptors) {
+                xmap.toXML(descriptor, root);
+            }
```
andreinechaev, line 140: 1. It would be nice if you can explain your logic here. 
You're trying to map all given descriptors to an XML element. But there is no verification that they are instances of the same class. It's an issue even in the old code - true. 

2. I guess, the advantage of using `Document` is to have an exception. 
3. `XMap#toXML(Object, Element)` is not clear on its intentions in terms of if it appends new tags or simply replaces

vpasquier: 1. The check is already done at the upper level

2. I dont get the exception part

3. Its adding children to the document `root`

The whole reason of this new code is to avoid having such contributions:

```
<xml....>
<root>
<thresolds>
    <threshold id="first"/>
<thresholds/>
</root>
<root>
<thresholds>
   <threshold id="second"/>
</thresholds>
</root>
```

The `root` is added in an hardcoded way of `XMAP#toXML(Object)` which is a bit toxic.

PR #283, opened 2020-11-10
AICORE-383: add s3 blob provider to helpers

andreinechaev:


PedroCardoso (approved):

PedroCardoso:
Do we have to backport to 10.10 also ?

PR #289, opened 2020-11-13
AICORE-391: Add GCP Image Properties Feature

michaelva:
This is a tentative to add the Image Properties feature from GCP. Not completely sure how to store the dominant colors returned by the API so for each I serialized the RGB code and pixel fraction as a label (string).


andreinechaev (approved):

andreinechaev:
Nice and clean. I think you might be interested in backporting it into 10.10. Thanks

PR #312, opened 2020-12-02
AICORE-404: add init audit operation for internal purpose

vpasquier:


nuxeo-ai-internal/pom.xml
```diff
@@ -0,0 +1,86 @@
...
+      <groupId>org.nuxeo.runtime</groupId>
+      <artifactId>nuxeo-runtime</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.nuxeo.ecm.automation</groupId>
+      <artifactId>nuxeo-automation-core</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.nuxeo.ecm.platform</groupId>
+      <artifactId>nuxeo-platform-audit-api</artifactId>
```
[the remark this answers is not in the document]
vpasquier: I cannot fetch this dependency - i let it as it is :/

nuxeo-ai-internal/src/main/java/org/nuxeo/ai/internal/InitAudit.java
```diff
@@ -0,0 +1,95 @@
...
+import org.nuxeo.ecm.core.api.DocumentModel;
+import org.nuxeo.ecm.core.api.NuxeoPrincipal;
+import org.nuxeo.ecm.platform.audit.api.AuditLogger;
+import org.nuxeo.ecm.platform.audit.api.ExtendedInfo;
+import org.nuxeo.ecm.platform.audit.api.LogEntry;
+import org.nuxeo.ecm.platform.audit.impl.ExtendedInfoImpl;
+import org.nuxeo.runtime.api.Framework;
+
+@Operation(id = ID, category = "AICore", label = "Init Audit", description = "Init audit for metrics")
+public class InitAudit {
```
PedroCardoso, line 38: IF the sole purpose of this operation is for tests, should it not be part of the test code ?

PedroCardoso (changes requested):

nuxeo-ai-internal/src/main/java/org/nuxeo/ai/internal/InitAudit.java
vpasquier: Yes it could but as we have in the nuxeo ai cloud repository, an internal module could be useful to integrate for migrating elements as well. We talked about it quickly with julien and it was not shocking to have a dedicated module for those kind of things

vpasquier: Actually I update my comment: this operation wont be used for unit testing - only for giving examples to dev on the front client side.

PR #342, opened 2021-01-20
AICORE-434: clean up nuxeo tags before saving

andreinechaev:


vpasquier:
And it needs a backport i guess too in 10.10 right?

PR #345, opened 2021-01-21
AICORE-295: adding metrics to ai core third party addons and insight

vpasquier:


addons/nuxeo-ai-aws-core/src/main/java/org/nuxeo/ai/rekognition/RekognitionServiceImpl.java
```diff
@@ -67,6 +68,14 @@
 
     protected volatile AmazonRekognition client;
 
+    protected AIComponent aiComponent;
+
+    @Override
+    public void activate(ComponentContext context) {
+        super.activate(context);
+        aiComponent = Framework.getService(AIComponent.class);
```
andreinechaev, line 76: the order of the component load should be after AIComponent then

jcarsique (approved):

addons/nuxeo-ai-aws-core/src/main/java/org/nuxeo/ai/comprehend/ComprehendServiceImpl.java
```diff
@@ -54,12 +65,16 @@ public DetectSentimentResult detectSentiment(String text, String languageCode) {
...
         DetectSentimentRequest request = new DetectSentimentRequest().withText(text).withLanguageCode(languageCode);
-        DetectSentimentResult result = getClient().detectSentiment(request);
 
-        if (log.isDebugEnabled()) {
-            log.debug("DetectSentimentResult is " + result);
+        try {
+            DetectSentimentResult result = getClient().detectSentiment(request);
+            if (log.isDebugEnabled()) {
+                log.debug("DetectSentimentResult is " + result);
+            }
```
PedroCardoso, line 73: Why do we need to have an if here ?

PedroCardoso: Same for all others.

addons/nuxeo-ai-aws-core/src/main/java/org/nuxeo/ai/comprehend/ComprehendServiceImpl.java
```diff
@@ -54,12 +65,16 @@ public DetectSentimentResult detectSentiment(String text, String languageCode) {
...
+            return result;
+        } finally {
+            awsMetrics.updateComprehendSentimentUnits(text.length() / 100L);
```
PedroCardoso, line 76: Should we measure it, even if there was a network errors, or configuration error ?

PedroCardoso (approved):

addons/nuxeo-ai-aws-core/src/main/java/org/nuxeo/ai/comprehend/ComprehendServiceImpl.java
vpasquier: It's something we're doing in Nuxeo platform too to avoid logging (even if the debug is not activated)

addons/nuxeo-ai-aws-core/src/main/java/org/nuxeo/ai/comprehend/ComprehendServiceImpl.java
vpasquier: Indeed, I have rewrite all of them to be skipped if there is an exception

addons/nuxeo-ai-aws-core/src/main/java/org/nuxeo/ai/rekognition/RekognitionServiceImpl.java
vpasquier: I set them now in start method of the components. Is it ok to do that without handling loading order?

vpasquier: Oh yes thats good: "Start the component. This method is called after all the components were resolved and activated"

andreinechaev (approved):

PR #346, opened 2021-01-25
AICORE-439: rename ai endpoint for rekognition

andreinechaev:


PedroCardoso:
I fail to understand how changing the endpoint name relates to the protocol changing.

PedroCardoso:
👍

PedroCardoso (approved):
Quick check that the https is not an issue, and should be good.

PR #352, opened 2021-02-01
AICORE-442: wrap ES audit queries

vpasquier:


andreinechaev (approved):
seems an overkill to configure freemaker template engine everytime, but it's not critical 👍

PR #366, opened 2021-02-22
AICORE-457: expose nuxeo.ai.insight.client.enable property and use it (backport 10.10)

andreacornaglia:
backport of https://github.com/nuxeo/nuxeo-ai/pull/364

PR #368, opened 2021-02-27
AICORE-457: remove webui contrib from mp config

vpasquier:


andreacornaglia (approved):

andreacornaglia (changes requested):
I think instead of just reverting this file we should either revert the whole PR https://github.com/nuxeo/nuxeo-ai/pull/366/commits or remove the walrus operator as Tiago recommended

PR #375, opened 2021-03-10
AICORE-459: JWT Alignment

vpasquier:
- align token generation
- cleanup old token
- load model service component after nuxeo client config
- assign managers group to nuxeo cloud client


nuxeo-ai-model/src/main/java/org/nuxeo/ai/cloud/NuxeoCloudClient.java
```diff
@@ -169,14 +171,8 @@ protected InsightClient configureClient(CoreSession session, @Nonnull CloudConfi
...
             claims.put(PublicClaims.SUBJECT, session.getPrincipal().getActingUser());
-            String[] groups = session.getPrincipal()
-                                     .getAllGroups()
-                                     .stream()
-                                     .filter(group -> group.startsWith(descriptor.projectId))
-                                     .toArray(String[]::new);
-            if (groups.length > 0) {
-                claims.put(NuxeoClaim.GROUP, groups);
-            }
+            String[] groups = { descriptor.projectId + "-managers" };
```
andreinechaev, line 174: you already have a constant for the postfix, please use it

andreinechaev (approved):
Overall there are a few things to note:
- There is extensive usage of Component Startup order, we should see if we can be startup agnostic. It means that: do we really need to execute any hardcoded actions during components' startups. Or there is a way to react on `applicationStarted` (if there is such) event. That would ensure that all components loaded. 
- Having user group verification in Cloud Client wrapper service is a fine double check approach, despite that I would consider it appropriately reacting on the cloud response. It would encapsulate user management and access on the cloud only. Of course, then we see that some dynamic configs still require group check, necessary evil, I assume

PR #381, opened 2021-03-18
AICORE-470: Fix duplicate captions - master2021

michaelva:


andreinechaev:
I'm closing the PR because we don't manually cherry pick on master 2021. It's an automatized process.

PR #387, opened 2021-03-19
AICORE-471: filter out versions in enrichments and no save on immutable

andreinechaev:


nuxeo-ai-pipes/src/main/java/org/nuxeo/ai/pipes/filters/NoVersionFilter.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+package org.nuxeo.ai.pipes.filters;
+
+import org.nuxeo.ecm.core.api.DocumentModel;
+
+public class NoVersionFilter implements Filter.DocumentFilter {
+
+    @Override
+    public boolean test(DocumentModel doc) {
+        return !doc.isVersion();
```
vpasquier, line 29: By anticipation of the next bug, I would add a check actually on isProxy too: proxies are published document that cannot be updated either (as far as i remember)

vpasquier (approved):

nuxeo-ai-pipes/src/main/java/org/nuxeo/ai/pipes/filters/NoVersionFilter.java
andreinechaev: we already have a filter for that kind https://github.com/nuxeo/nuxeo-ai/blob/master/nuxeo-ai-pipes/src/main/java/org/nuxeo/ai/pipes/filters/NotSystemOrProxyFilter.java

jcarsique (approved):

PR #437, opened 2021-06-14
AICORE-503: fixing default repo name

andreinechaev:


andreinechaev:
Was applied while getting rid of sprint branches

PR #439, opened 2021-06-15
AICORE-502: fixing translation+better logging

andreinechaev:


jcarsique (approved):

addons/nuxeo-ai-aws-core/src/test/java/org/nuxeo/ai/TestTranslateService.java
```diff
@@ -55,7 +56,7 @@ public void testTranslate() {
 
         assertTranslation("aws.translate.en_es", "I am very disappointed", "Estoy muy decepcionado");
         assertTranslation("aws.translate.en_pt", "I am very happy", "Estou muito feliz");
-        assertTranslation("aws.translate.en_fr", "I am very happy", "Je suis très heureux");
+        assertTranslation("aws.translate.en_fr", "I am very happy", "Je suis très ");
```
PedroCardoso, line 59: Tomorrow they will update the model and it will return "je suis super content" :) 
anyway, not a blocker.

PedroCardoso (approved):

addons/nuxeo-ai-aws-core/src/test/java/org/nuxeo/ai/TestTranslateService.java
andreinechaev: I was hoping to exclude the diff on synonyms :) in the ideal world we should not test the outcome of a 3rd party service. The test should check the ability to call and store. But of course it is open for debate

PedroCardoso: Agree with you, still I am fine with what it is now.

PR #442, opened 2021-06-15
AICORE-502: fixing translation+better logging

andreinechaev:


andreinechaev:
No need since no sprint branches. Patches Applied in #440 and #438

PR #459, opened 2021-08-02
AICORE-519: add custom datasource claim to authenticate via jwt

vpasquier:


jcarsique (approved):

nuxeo-ai-model/src/main/java/org/nuxeo/ai/listeners/InvalidateModelDefinitionsListener.java
```diff
@@ -112,9 +112,9 @@ public void handleEvent(EventBundle bundle) {
                 all = Collections.emptyMap();
                 log.warn("No active models were found");
             }
-        } catch (IOException e) {
-            log.error(e);
-            throw new NuxeoException(e);
+        } catch (IOException | CacheLoader.InvalidCacheLoadException e) {
+            log.warn("Cannot retrieve published models from Insight: {}", e.getMessage());
+            return;
```
andreinechaev, line 117: are you sure that swallowing an exception is the most optimal way?

andreinechaev (approved):

nuxeo-ai-model/src/main/java/org/nuxeo/ai/listeners/InvalidateModelDefinitionsListener.java
vpasquier: This is what we were talking about last time with you and Julien. That we didn't need to see all those exceptions when a user is not authenticated. I can let it as it is if you changed your mind or if you have another idea?

vpasquier: Talked with andrei, we gonna find another way

PR #461, opened 2021-08-15
AICORE-534: Don't store metrics service

michaelva:


andreinechaev (approved):
Please cherry pick into master-10.10

PR #462, opened 2021-08-16
AICORE-535: init crophint call counter

michaelva:


andreinechaev (approved):
please cherry-pick into master-10.10

PR #480, opened 2021-11-16
AICORE-517: find similar operation

andreinechaev:


nuxeo-ai-similar-content/src/main/java/org/nuxeo/ai/similar/content/operation/FindSimilar.java
```diff
@@ -0,0 +1,165 @@
...
+
+    protected List<String> findSimilarIds(@Nullable DocumentModel doc, @Nullable Blob blob) throws OperationException {
+        Optional<InsightClient> client = this.client.getClient(session);
+        if (client.isEmpty()) {
+            throw new OperationException(
+                    "Could not obtain Insight Client for user " + session.getPrincipal().getActingUser());
+        }
+
+        String id = doc == null ? null : doc.getId();
+        TensorInstances tensor = constructTensor((ManagedBlob) blob, id);
```
vpasquier, line 115: Something to discuss: Here we can have a null blob  (so a null tensor instance) but at the same time, we are having the xpath so we could get a potential blob from there. On the other operations im doing, i don't take any blob as input, always the doc/xpath. Something maybe to sync together.

nuxeo-ai-similar-content/src/main/java/org/nuxeo/ai/similar/content/operation/FindSimilar.java
```diff
@@ -0,0 +1,165 @@
...
+        if (blob == null) {
+            return null;
+        }
+
+        Map<String, TensorInstances.Tensor> props = new HashMap<>();
+        props.put(xpath, TensorInstances.Tensor.image(convertImageBlob(blob)));
+        return new TensorInstances(id, singletonList(props));
+    }
+
+    protected String convertImageBlob(ManagedBlob blob) {
```
vpasquier, line 156: Something to discuss: Here we need to use the same conversion for both index/find operations. In my ticket we are taking about conversion like we did for the export `org.nuxeo.ai.pipes.functions.PropertyUtils#getPictureConversion`. But with different specifications which is having less than 5MB in size (no restriction in term of dimension). Something maybe to sync too.

vpasquier (commented):
I have some questions in this PR that will impact mine as well.

PedroCardoso (approved):

vpasquier (approved):

PR #481, opened 2021-11-17
AICORE-549: Handle async result pagination

michaelva:


addons/nuxeo-ai-aws-core/src/main/java/org/nuxeo/ai/enrichment/async/DetectCelebritiesEnrichmentProvider.java
```diff
@@ -84,16 +89,32 @@ public void init(EnrichmentDescriptor descriptor) {
...
+
+            List<AIMetadata.Tag> currentPageTags = result.getCelebrities()
+                                             .stream()
+                                             .map(c -> newCelebrityTag(c.getCelebrity(), c.getTimestamp()))
+                                             .filter(Objects::nonNull)
+                                             .collect(Collectors.toList());
+
+            tags.addAll(currentPageTags);
+            nativeCelebrityObjects.addAll(result.getCelebrities());
+        } while (result.getNextToken() != null);
```
andreinechaev, line 114: it makes me wonder what is the max size of the outcome list. Should it be limited by user's configuration? Should we consider Set to enforce uniqueness?

michaelva: I thought about it and IMO the notion of max size doesn't really make any sense for a video. A 1h long video will have more results than a 10s one most of the time. Uniqueness should already be enforced for a given timestamp.

andreinechaev: my concern is that this data might be stored into nuxeo tags if the related feature is enabled. Won't it affect the system and UI if we'll have 1000s of tags?

michaelva: OOTB there is no feature to store video rekognition labels as tags. It's doesn't follow the enrichment pipe logic after the request to rekognition. The way results are handled is hardcoded in the listeners. In this case "save as facet".

andreinechaev (approved):

andreinechaev:
please backport on 10.10. Thank you!

PR #482, opened 2021-11-17
AICORE-527: add dedup listener

vpasquier:


vpasquier:
For info the listener is sync because of beforeModification event that requires it. We could separate, let me know.

vpasquier:
Also i still have to fix a mockito issue i have to validate the unit tests - do not merge

andreinechaev (approved):

PR #487, opened 2021-11-25
AICORE-551: index repository BAF impl

andreinechaev:

PR #488, opened 2021-11-30
Task aicore 553 index baf op

andreinechaev:


andreinechaev:
only last commit needs to be reviewed - the rest is in #487

nuxeo-ai-similar-content/src/main/java/org/nuxeo/ai/similar/content/services/SimilarServiceComponent.java
```diff
@@ -84,6 +113,91 @@ public String getOperationID() {
...
+                                   .map(BulkCodecs.getStatusCodec()::decode)
+                                   .filter(status -> INDEX_ACTION_NAME.equals(status.getAction()))
+                                   .anyMatch(status -> status.getState().equals(BulkStatus.State.RUNNING)
+                                           || status.getState().equals(BulkStatus.State.SCHEDULED));
+        if (alreadyRunning) {
+            throw new ConcurrentUpdateException("Only one deduplication index action is allowed at a time");
+        }
+
+        if (!reindex) {
+            query += " AND ecm:mixinType != " + NXQL.escapeString(DEDUPLICATION_FACET);
```
vpasquier, line 131: Should we not have this check on the query for the facet filter for all use cases?

vpasquier (approved):
Just a question but all good!

nuxeo-ai-similar-content/src/main/java/org/nuxeo/ai/similar/content/services/SimilarServiceComponent.java
andreinechaev: we can indeed check if such expression already takes place and apply it optionally. If initial query wasn't properly structured and contains this condition - it's a considered as a human error IMO

PR #489, opened 2021-11-30
Task aicore 552 index progress op

andreinechaev:


andreinechaev:
only last commit needs to be reviewed, the rest of work can be found in #487 and #488

vpasquier (approved):

PR #492, opened 2021-12-02
AICORE-528: duplications found event on create/update

andreinechaev:


nuxeo-ai-similar-content/src/main/java/org/nuxeo/ai/similar/content/services/SimilarContentService.java
```diff
@@ -72,6 +75,20 @@
      */
     boolean index(DocumentModel doc, String xpath) throws IOException;
 
+    List<DocumentModel> findSimilar(CoreSession session, DocumentModel doc, String xpath) throws IOException;
```
vpasquier, line 78: javadoc

vpasquier (approved):
I would put the event constants in DedupConstants to gather them at the same place and set in the upgrade notes of the jira ticket that we have an event `similarDocumentsFound` that can be used for the customer custom implem (in order for us to remember what should be documented with the dedup in general). To discuss of course and otherwise all good!

PR #499, opened 2021-12-08
AICORE-555: consistent operation names

andreinechaev:


andreinechaev:
merged without review due to agreed on ticket with provided names

PR #527, opened 2022-09-02
AICORE-572: Add GCP SafeSearch

ThibArg:
Adding GCP safe search. Main goals are:
- Show we connect to several providers (already the case, but adding feature for one of them)
- Allow for feature comparison (for example, for SafeSearch, AWS returns more details), god to explain choosing the right provider for your data is important
- Explain HxInsight would be even better if rules/data are very specific and need training because aws/google never return correct tagging.


andreinechaev (approved):
Nice Image provided ;) 
Overall nice and clean contribution, thanks!

PR #534, opened 2022-10-16
task AICORE 579 release publish

vpasquier:


.github/workflows/release_publish.yaml
```diff
@@ -0,0 +1,189 @@
...
+  release-publish-10.10:
+    name: 'Release Publish Master 10.10'
+    runs-on: ubuntu-latest
+    if: github.ref == 'refs/heads/master-10.10'
+    env:
+      MAVEN_OPTS: "-Xms1g -Xmx1536m"
+    outputs:
+      version: ${{ steps.project.outputs.version }}
+    steps:
+    - uses: actions/checkout@v3
```
andreinechaev, line 26: don't you need a branch name here. It's a manual workflow if I understand correctly, so it requires more specification

.github/workflows/release_publish.yaml
```diff
@@ -0,0 +1,189 @@
...
+
+    - name: Post Release
+      run: |
+        git checkout ${{ github.ref }}
+        find ./ -type f -exec sed -i '' -e "s/${{ github.event.inputs.version }}-SNAPSHOT/${{ steps.bump_version.outputs.next-version }}-SNAPSHOT/" {} \;
+        git add .
+        git commit -m "Post Release ${{ github.event.inputs.version }}"
+        git push ${{ github.ref }}
+
+  release-publish-lts:
```
andreinechaev, line 104: Seems that it is identical to 10.10 the only difference is java version, right? It might be one of the default parameters . It will also let us release from a branch

.github/workflows/release_publish.yaml
vpasquier: its still in test

PR #536, opened 2022-10-24
AICORE-585: fix tmp file creation

andreinechaev:


nuxeo-ai-model/src/main/java/org/nuxeo/ai/bulk/RecordWriterBatchComputation.java
```diff
@@ -82,5 +82,14 @@ public void batchFailure(ComputationContext context, String inputStream, List<Re
         String commandId = records.isEmpty() ? "" : records.get(0).getKey();
         log.warn("Batch failure \"{}\" batch of {} records with command {}.", metadata.name(), records.size(),
                 commandId);
+        Codec<ExportRecord> codec = getAvroCodec(ExportRecord.class);
+        log.warn("Mark as failed {} Export Records for command ID {}", records.size(), commandId);
+        records.forEach(rec -> {
+            ExportRecord decoded = codec.decode(rec.getData());
+            decoded.setFailed(true);
+            byte[] encoded = codec.encode(decoded);
```
jjliggett, line 90: Is this value different from the value `rec.getData()` (which gets decoded on line 88)?

jjliggett: We discussed this in a call. This encoding is required, because we modified the value in the previous line (line 89) - `decoded.setFailed(true)`. We need to encode the result after that modification.

jjliggett (approved):

PR #541, opened 2022-11-02
AICORE-589: no empty outputs during export

andreinechaev:


vpasquier (approved):
All good except tests (assertj version?)

PR #551, opened 2023-01-17
AICORE-596: bulk status for indexing stored in kvs

andreinechaev:


nuxeo-ai-similar-content/src/main/java/org/nuxeo/ai/similar/content/services/SimilarServiceComponent.java
```diff
@@ -153,28 +171,46 @@ public String index(CoreSession session, String query, boolean reindex) throws I
...
-        return kv.keyStream(STATUS_PREFIX)
-                 .map(kv::get)
-                 .map(BulkCodecs.getStatusCodec()::decode)
-                 .filter(status -> INDEX_ACTION_NAME.equals(status.getAction()))
-                 .filter(status -> id == null || status.getId().equals(id))
-                 .sorted(Comparator.comparing(BulkStatus::getState))
-                 .map(BulkProgressStatus::new)
-                 .findFirst()
-                 .orElse(null);
+
```
bdelbosc, line 201: just use `bs.getStatus(id)` instead of kvs

PR #597, opened 2023-04-26
AICORE-612: Check for doc base versions when enriching & improve logs

jjliggett:
https://jira.nuxeo.com/browse/AICORE-612

We have encountered a NullPointerException within SaveEnrichmentFunction. This was a result of corrupted documents where the documents did not have base versions. This change makes an improvement to the class. We now check if the base version is null and if it is null and the document is not checked out, we exit, returning null. Additionally, this change improves logging within the class, adding an error log for unexpected exceptions along with a couple other logs. The class also had a couple small refactors as well.

```
java.lang.NullPointerException: null at org.nuxeo.ecm.core.versioning.StandardVersioningService.doCheckOut(StandardVersioningService.java:374) ~[nuxeo-core-10.10-HF71.jar:?] at org.nuxeo.ecm.core.versioning.StandardVersioningService.doPreSave(StandardVersioningService.java:325) ~[nuxeo-core-10.10-HF71.jar:?] at org.nuxeo.ecm.core.versioning.VersioningComponent.doPreSave(VersioningComponent.java:423) ~[nuxeo-core-10.10-HF71.jar:?] at org.nuxeo.ecm.core.api.AbstractSession.saveDocument(AbstractSession.java:1613) ~[nuxeo-core-10.10-HF71.jar:?] at org.nuxeo.ai.functions.SaveEnrichmentFunction.lambda$null$0(SaveEnrichmentFunction.java:49) ~[nuxeo-ai-core-2.7.19.jar:?] at org.nuxeo.ecm.core.api.CoreInstance$1.run(CoreInstance.java:186) ~[nuxeo-core-api-10.10-HF71.jar:?]
```

### DIFF
--- a/nuxeo-ai-core/src/main/java/org/nuxeo/ai/functions/SaveEnrichmentFunction.java
+++ b/nuxeo-ai-core/src/main/java/org/nuxeo/ai/functions/SaveEnrichmentFunction.java
@@ -25,6 +25,7 @@ import org.nuxeo.ai.enrichment.EnrichmentMetadata;
 import org.nuxeo.ai.services.DocMetadataService;
 import org.nuxeo.ecm.core.api.CoreInstance;
 import org.nuxeo.ecm.core.api.DocumentModel;
+import org.nuxeo.ecm.core.api.DocumentRef;
 import org.nuxeo.ecm.core.api.validation.ConstraintViolation;
 import org.nuxeo.ecm.core.api.validation.DocumentValidationException;
 import org.nuxeo.runtime.api.Framework;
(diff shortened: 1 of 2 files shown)


nuxeo-ai-core/src/main/java/org/nuxeo/ai/functions/SaveEnrichmentFunction.java
```diff
@@ -41,30 +42,52 @@ public class SaveEnrichmentFunction extends AbstractEnrichmentConsumer {
...
+                log.warn("Failed to save enrichment for document {}.", metadata.context.documentRef);
+                return null;
+            }
+
+            log.debug("Checking if the document is checked out and if a base version exists for the document {}.",
+                    doc.getId());
+
+            DocumentRef baseVersionRef = session.getBaseVersion(doc.getRef());
+
+            if (baseVersionRef == null && !doc.isCheckedOut()) {
```
jjliggett, line 59: The `!doc.isCheckedOut()` check was added in dd3b862658a35f863a8a0ab7b85739af3a58b189 (see -  https://github.com/nuxeo/nuxeo-ai/compare/14648a99f3c449402bab08e28f3c4c6c78495c30..dd3b862658a35f863a8a0ab7b85739af3a58b189 ), because there are valid cases where documents do not have base versions (such as a document that was never versioned). The check before was too strong and was causing one of our tests to fail, as shown below.

This change relates to the code path from the exception where a document is checked in and the versioning service tries to check out the document but fails as there isn't a base version:

https://github.com/nuxeo/nuxeo-lts/blob/eb3a6cda7c8a5a7d9e212e105a76bee1bb801166/modules/core/nuxeo-core/src/main/java/org/nuxeo/ecm/core/versioning/StandardVersioningService.java#L368-L385

For a description of the `isCheckedOut()`, please see here:

https://github.com/nuxeo/nuxeo-lts/blob/eb3a6cda7c8a5a7d9e212e105a76bee1bb801166/modules/core/nuxeo-core-api/src/main/java/org/nuxeo/ecm/core/api/DocumentModel.java#L351-L362

The test failure was the following:

```
java.lang.AssertionError: The document must have the enrichment facet
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.nuxeo.ai.enrichment.TestConfiguredStreamProcessors.testConfiguredStreamProcessor(TestConfiguredStreamProcessors.java:135)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
```

( https://github.com/nuxeo/nuxeo-ai/pull/597/checks?check_run_id=13046678534 )

jjliggett:
We ran into a couple flaky errors within the workflow, first `BulkEnrichmentTest.shouldNotBacktrackConfirmedValue` and then `AnnotationConfirmTest.shouldNotBacktrackConfirmedValue`. The third attempt passed - https://github.com/nuxeo/nuxeo-ai/actions/runs/4820357966. I don't think this change caused this flakyness, but it's something we should improve in the future.

nuxeo-ai-core/src/main/java/org/nuxeo/ai/functions/SaveEnrichmentFunction.java
```diff
@@ -41,30 +42,52 @@ public class SaveEnrichmentFunction extends AbstractEnrichmentConsumer {
...
+                log.warn(
+                        "Failed to save enrichment. The document is not checked out and no base version was found for document {}.",
```
bdelbosc, line 61: You could state that the document is corrupted and require a manual intervention to be fixed, this is not something to be expected, could be better at error level.

nuxeo-ai-core/src/main/java/org/nuxeo/ai/functions/SaveEnrichmentFunction.java
```diff
@@ -41,30 +42,52 @@ public class SaveEnrichmentFunction extends AbstractEnrichmentConsumer {
...
+                        doc.getId());
+                return null;
+            }
+
+            if (doc.isImmutable()) {
+                log.error("Attempt to write into an Immutable Document Model id: {}, AI Model name {}", doc.getId(),
+                        metadata.getModelName());
```
bdelbosc, line 68: It will be interesting to understand why we are trying to update an immutable version or proxy ? 
This should be filtered upstream instead of computing enrichment that cannot be persisted.

bdelbosc (approved):
Your github comments are valuable and should be in the Jira ticket so they don't get lost.

nuxeo-ai-core/src/main/java/org/nuxeo/ai/functions/SaveEnrichmentFunction.java
jjliggett: This has been changed in [f0c7c09](https://github.com/nuxeo/nuxeo-ai/commit/f0c7c09a20df8eeb38f0a4efb2a42cae1a10abc6) - https://github.com/nuxeo/nuxeo-ai/compare/dd3b862658a35f863a8a0ab7b85739af3a58b189..f0c7c09a20df8eeb38f0a4efb2a42cae1a10abc6

nuxeo-ai-core/src/main/java/org/nuxeo/ai/functions/SaveEnrichmentFunction.java
jjliggett: Thanks, @bdelbosc! Filtering upstream sounds like a good idea for a future enhancement. I think this could be a good additional check here though, so we could keep this code even with filtering.

bdelbosc (approved):

jjliggett:
> Again I am in favor of fixing the corrupted documents at the db level instead of adapting the code to handle them because they are exceptional and caused by a bug which is now fixed. Your github's comments are valuable and should be in the Jira ticket so they don't get lost.

Thanks, @bdelbosc! I'll add the comments to Jira. Fixing the documents at the database level would be good, but this change will also be beneficial as we this error is unrecoverable here - any retries saving enrichments with this error would fail. We want to skip retrying those records and want to have reporting and monitoring of such errors instead.

jjliggett:
The workflow succeeded! Merging now.

image

( https://github.com/nuxeo/nuxeo-ai/actions/runs/4831429221/jobs/8608875464 )

PR #598, opened 2023-04-28
Release 2.7.20

jjliggett:
This pull request will publish release 2.7.20 and will later be used to prepare v2.7.21-SNAPSHOT with a second commit.

This includes the changes in #597

The previous release PR was Release 2.7.19 - https://github.com/nuxeo/nuxeo-ai/pull/596

PR #599, opened 2023-05-01
Post release 2.7.20

jjliggett:
This is a follow-up to https://github.com/nuxeo/nuxeo-ai/pull/598

It is similar to https://github.com/nuxeo/nuxeo-ai/pull/596

### DIFF
--- a/addons/nuxeo-ai-aws-core/pom.xml
+++ b/addons/nuxeo-ai-aws-core/pom.xml
@@ -4,7 +4,7 @@
   <parent>
     <groupId>org.nuxeo.ai</groupId>
     <artifactId>nuxeo-ai-addons</artifactId>
-    <version>2.7.20</version>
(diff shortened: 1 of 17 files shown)


ryewilson:
There is a way to configure maven poms so that the version need only be specified in the root pom and it cascades down to all child poms. Unless there's a specific reason that was not done here, I will attempt to implement this so that future releases touch fewer files.

jjliggett:
> There is a way to configure maven poms so that the version need only be specified in the root pom and it cascades down to all child poms. Unless there's a specific reason that was not done here, I will attempt to implement this so that future releases touch fewer files.

Hi, @ryewilson, sounds great! I'm not aware of a specific reason why that wasn't done here. We should always keep the packages at the same version within our codebase, so I think that would be an ideal change.

CodySalyi (approved):

PR #600, opened 2023-05-02
AICORE-613: Delete unused Jenkinsfile

jjliggett:
We now use GitHub Actions. Kevin mentioned that the Jenkins instance was destroyed.


ryewilson (approved):
👍

PR #604, opened 2023-05-04
AICORE-616: Resolve error in ComprehendService, skipping analysis beyond quotas

jjliggett:
AICORE-616: Resolve error in ComprehendService, skipping analysis beyond quotas

We utilize Amazon Comprehend within our ComprehendService. There are quota limits that are enforced when detecting entities, key-phrases, dominant languages, sentiment, targeted sentiment, and syntax. Previously, if we tried to analyze text with a text length beyond the quota, we would encounter a TextSizeLimitExceededException. This change skips Amazon Comprehend analysis when text is larger than the specified quotas. In the future, we may implement another solution.

The guidelines and quotas for Amazon Comprehend can be found here: https://docs.aws.amazon.com/comprehend/latest/dg/guidelines-and-limits.html

PR #603, opened 2023-05-03
AICORE-615: Fix workflow package upload issue

jjliggett:
This is a quick fix related to https://github.com/nuxeo/nuxeo-ai/pull/602

image

PR #607, opened 2023-05-05
Forward port from 2.7.20 and 2.7.21

jjliggett:
This is similar to the following earlier forward port: https://github.com/nuxeo/nuxeo-ai/pull/592

This forward port includes 3 commits:

AICORE-612: Check for base versions when enriching & improve logs

We have encountered a NullPointerException within SaveEnrichmentFunction.
This was a result of corrupted documents where the documents did not have
base versions. This change makes an improvement to the class. We now
check if the base version is null and if it is null and the document is
not checked out, we exit, returning null in the function. Additionally,
this change improves logging within the class, adding an error log for
unexpected exceptions along with a couple other logs. The class also had
a couple small refactors as well.

This is a forward port from master-10.10. The code is mostly the same
as the original fix with minor changes to SaveEnrichmentFunction
related to ValidationViolation. The original change can be found in
commit f0c7c09 and pull request #597.

AICORE-614: Deploy packages to Nexus and the Marketplace
This is a forward port from master-10.10. The original commit is
0c4e12a and its relevant pull request was #602. This forward port also
includes a small fix provided in AICORE-615 where a copy-paste issue
in the previous commit was corrected. That fix has been included in this
forward port commit. For reference to the original quick fix, see the
commit 50e2091 and pull request #603.

AICORE-616: Resolve ComprehendService bug, skip text over quotas
We utilize Amazon Comprehend within our ComprehendService. There are
quota limits that are enforced when detecting entities, key-phrases,
dominant languages, sentiment, targeted sentiment, and syntax.
Previously, if we tried to analyze text with a text length beyond the
quota, we would encounter a TextSizeLimitExceededException. This change
skips Amazon Comprehend analysis when text is larger than the specified
quotas. In the future, we may implement another solution.

This is a forward port from master-10.10. For reference, see commit
6e84c5c and pull request #604.


ryewilson (approved):
👍

PR #608, opened 2023-05-05
Use {revision} to consolidate version

ryewilson:
By using the revision property this allows us to centralize where the version is specified to just the parent pom. All child poms will use the ${revision} property, which resolves to the value set in the root.

### DIFF
--- a/addons/nuxeo-ai-aws-core/pom.xml
+++ b/addons/nuxeo-ai-aws-core/pom.xml
@@ -4,7 +4,7 @@
   <parent>
     <groupId>org.nuxeo.ai</groupId>
     <artifactId>nuxeo-ai-addons</artifactId>
-    <version>3.5.10-SNAPSHOT</version>
+    <version>${revision}</version>
   </parent>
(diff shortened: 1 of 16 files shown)


jjliggett:
Hi, @ryewilson, can you please create an AICORE Jira card for this change and then include that in the commit message and pull request title?

ryewilson:
> Hi, @ryewilson, can you please create an AICORE Jira card for this change and then include that in the commit message and pull request title?

I believe I'm going to close this PR. The changes are not building correctly due to a bug in an outdated version of a maven plugin (maven-enforcer-plugin v1.4.1) defined in the parent pom that we have no control over.

jjliggett:
> > Hi, @ryewilson, can you please create an AICORE Jira card for this change and then include that in the commit message and pull request title?
> 
> I believe I'm going to close this PR. The changes are not building correctly due to a bug in an outdated version of a maven plugin (maven-enforcer-plugin v1.4.1) defined in the parent pom that we have no control over.

Okay, that's unfortunate. Thank you for looking into this and for sharing this update, @ryewilson!!

PR #609, opened 2023-05-18
AICORE-618: Add deploy failure slack notifications

jhunger:
This will send a slack notification on deployment failures.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -35,7 +35,7 @@ on:
 
 jobs:
   build:
-    name: 'Build 10.10'
+    name: "Build 10.10"
     runs-on: ubuntu-latest
     if: ${{ github.base_ref == 'master-10.10' || github.ref_name == 'master-10.10' }}
     env:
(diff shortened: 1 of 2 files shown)


.github/workflows/pipeline_release.yml
```diff
@@ -7,106 +7,119 @@ on:
...
-        git checkout ${{ github.base_ref }}
-        find ./ -type f -exec sed -i '' -e "s/$CURRENT_VERSION/${{ steps.bump_version.outputs.next-version }}/" {} \;
-        git clean -f -d
-        git add .
-        git commit -m "Post Release $CURRENT_VERSION"
-        git push origin ${{ github.base_ref }}
+  notify-on-error:
+    runs-on: ubuntu-latest
+    needs: release-publish
+    if: ${{ failure() && github.event_name == 'push' }}
```
jjliggett, line 116: This workflow is triggered by a manual `workflow_dispatch`

https://github.com/nuxeo/nuxeo-ai/blob/63930dbee31bd633c43c4b1eca91802619888bca/.github/workflows/pipeline_release.yml#L4

I think the event name wouldn't be push. Can you remove the `github.event_name == 'push'` condition? Since this is always manually run, we can alert on all workflow failures for this workflow.

.github/workflows/build_and_test.yml
```diff
@@ -35,220 +35,234 @@ on:
...
+        if: ${{ inputs.should-push }}
+        env:
+          PACKAGE: addons/nuxeo-ai-image-quality-package/target/nuxeo-ai-image-quality-package-${{ env.VERSION }}.zip
+          CONNECT_URL: https://connect.nuxeo.com/nuxeo
+        run: curl -i -u "${{ secrets.connect-auth-user }}:${{ secrets.connect-auth-token }}" -F package=@$PACKAGE "$CONNECT_URL/site/marketplace/upload?batch=true"
+
+  notify-on-error:
+    runs-on: ubuntu-latest
+    needs: [build, build-master]
+    if: ${{ failure() }}
```
jjliggett, line 259: Since the "CI for PR" (`pipline_pr.yaml`) workflow calls this workflow for pull requests and pushes, this could alert on pull requests now.

https://github.com/nuxeo/nuxeo-ai/blob/63930dbee31bd633c43c4b1eca91802619888bca/.github/workflows/pipeline_pr.yaml#L1-L6

https://github.com/nuxeo/nuxeo-ai/blob/63930dbee31bd633c43c4b1eca91802619888bca/.github/workflows/pipeline_pr.yaml#L9-L14

Could you update this condition to also check for the `should-push` input or as an alternative, could you move this job to the `pipeline_pr` workflow and check that the event is "push"?

```suggestion
    if: ${{ failure() && inputs.should-push }}
```

https://github.com/nuxeo/nuxeo-ai/blob/8fb7348bfc2bd715eab99050df7f1de378b737f9/.github/workflows/build_and_test.yml#L16-L19

Either of those two changes would make it so pull request workflow failures wouldn't notify our team while we would get notifications of release failures.

jjliggett (approved):
Thanks for the changes; looks good!

Can you also cherry-pick this and open a PR for the master-10.10 branch as well, for builds for Nuxeo 10.10?

https://github.com/nuxeo/nuxeo-ai/tree/master-10.10

PR #610, opened 2023-05-23
AICORE-618: Add deploy failure slack notifications

jhunger:
Added a slack failure notification.


.github/workflows/pipeline_pr.yaml
```diff
@@ -18,4 +18,17 @@ jobs:
       packages-auth-user: ${{secrets.PACKAGES_AUTH_USER}}
       packages-auth-token: ${{secrets.PACKAGES_AUTH_TOKEN}}
       connect-auth-user: ${{secrets.CONNECT_AUTH_USER}}
-      connect-auth-token: ${{secrets.CONNECT_AUTH_TOKEN}}
\ No newline at end of file
+      connect-auth-token: ${{secrets.CONNECT_AUTH_TOKEN}}
+  notify-on-error:
```
jjliggett, line 22: Since the job before this job calls the `build_and_test` workflow, we should only keep the error notification in this workflow or the other workflow. The configurations look good in both, but if we have both implemented, then we would probably get notified twice since there would be a failure in the `build_and_test` workflow which would also cause a failure in this workflow as well

jjliggett (approved):
Looks good! Thanks, @jhunger! It seems we ran into a flaky test failure. I'll rerun that, but it shouldn't be a concern for this PR.

"BulkEnrichmentTest.shouldNotBacktrackConfirmedValue"

https://github.com/nuxeo/nuxeo-ai/pull/610/checks?check_run_id=13723608840

PR #633, opened 2023-08-03
AICORE-626: Make NuxeoCloudClient dataset uploads more resilient (Forward port from 2.7.23 to 2023)

jjliggett:
We have encountered a NullPointerException within the NuxeoCloudClient, when creating and uploading dataset batches. A blob file was null, causing the program to run into into a NullPointerException within the FileBlob constructor. This commit makes the NuxeoCloudClient more resilient. We now check for null blob files within the createBatch() and uploadDataset() methods.

https://jira.nuxeo.com/browse/AICORE-626

This is identical to the following fix for 10.10 and 2021:

- https://github.com/nuxeo/nuxeo-ai/pull/627
- https://github.com/nuxeo/nuxeo-ai/pull/629

### DIFF
--- a/nuxeo-ai-model/src/main/java/org/nuxeo/ai/cloud/NuxeoCloudClient.java
+++ b/nuxeo-ai-model/src/main/java/org/nuxeo/ai/cloud/NuxeoCloudClient.java
@@ -294,6 +294,12 @@ public class NuxeoCloudClient extends DefaultComponent implements CloudClient {
 
     protected boolean createBatch(BatchUpload batchUpload, String name, String index, Blob blob) {
         if (blob == null) {
+            log.warn("Batch blob for {} is null", name);
+            return false;
+        }
+
+        if (blob.getFile() == null) {
+            log.warn("Could not get blob file for {}", name);
             return false;
         }
 
@@ -314,9 +320,9 @@ public class NuxeoCloudClient extends DefaultComponent implements CloudClient {
         Blob evalData = (Blob) dataset.getPropertyValue(DATASET_EXPORT_EVALUATION_DATA);
         Blob statsData = (Blob) dataset.getPropertyValue(DATASET_EXPORT_STATS);
 
-        if ((trainingData == null || trainingData.getLength() == 0) && (evalData == null
-                || evalData.getLength() == 0)) {
-            log.warn("Job/Command: {} has no training and/or evaluation data. Document {}", jobId, dataset.getId());
+        if ((trainingData == null || trainingData.getLength() == 0 || trainingData.getFile() == null) && (
+                evalData == null || evalData.getLength() == 0 || evalData.getFile() == null)) {
+            log.warn("Job/Command: {} has neither training nor evaluation data. Document {}", jobId, dataset.getId());
         } else if (statsData == null || statsData.getLength() == 0) {
             log.warn("Job/Command: {} has no statistics data.", jobId);
         } else {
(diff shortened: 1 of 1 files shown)


ryewilson (approved):

nuxeo-ai-model/src/main/java/org/nuxeo/ai/cloud/NuxeoCloudClient.java
```diff
@@ -314,8 +320,8 @@ public String uploadDataset(@Nonnull DocumentModel dataset) {
         Blob evalData = (Blob) dataset.getPropertyValue(DATASET_EXPORT_EVALUATION_DATA);
         Blob statsData = (Blob) dataset.getPropertyValue(DATASET_EXPORT_STATS);
 
-        if ((trainingData == null || trainingData.getLength() == 0) && (evalData == null
-                || evalData.getLength() == 0)) {
+        if ((trainingData == null || trainingData.getLength() == 0 || trainingData.getFile() == null) && (
+                evalData == null || evalData.getLength() == 0 || evalData.getFile() == null)) {
             log.warn("Job/Command: {} has no training and/or evaluation data. Document {}", jobId, dataset.getId());
```
nichiporenko: I assume we should update the log message (remove the `/or` part) since this if statement is true only if both training and evaluation data are not present. So having `... has no training AND evaluation data ...` in the log makes more sense, at least for me right now

nuxeo-ai-model/src/main/java/org/nuxeo/ai/cloud/NuxeoCloudClient.java
```diff
@@ -330,13 +336,34 @@ public String uploadDataset(@Nonnull DocumentModel dataset) {
...
+                    batch1 = null;
+                } else if (trainingData.getLength() == 0) {
+                    log.warn("Job/Command: {} has no training data - data has 0 length", jobId);
+                    batch1 = null;
+                } else if (trainingData.getFile() == null) {
+                    log.warn("Job/Command: {} has no training data - data blob file is null", jobId);
+                    batch1 = null;
+                } else {
+                    batch1 = createBatch(batchUpload, "training", "0", trainingData) ? batch1 : null;
+                }
```
nichiporenko, line 338: Maybe just to have smaller methods, for better readability, we can extract methods from lines [338 to 351](https://github.com/nuxeo/nuxeo-ai/pull/633/files#diff-2b131e904e91d058146eb3ee146b3d0ef9146b916007f599391bdf34e36db618R338-R351) and from [353 to 366](https://github.com/nuxeo/nuxeo-ai/pull/633/files#diff-2b131e904e91d058146eb3ee146b3d0ef9146b916007f599391bdf34e36db618R353-R366) (including the lines with the assignments to `batchUpload` variable like this `batchUpload = client.getBatchUpload(CHUNK_100_MB);`)

And set the names for methods: `createTrainingDataBatch()`, `createEvaluationDataBatch()` (for example)

nuxeo-ai-model/src/main/java/org/nuxeo/ai/cloud/NuxeoCloudClient.java
jjliggett: Thanks, @nichiporenko! I'm updating this to "has neither training nor evaluation data"

nuxeo-ai-model/src/main/java/org/nuxeo/ai/cloud/NuxeoCloudClient.java
jjliggett: Thank you, @nichiporenko! I think that's a good idea. Pushing a change for this now.

jjliggett: This has been addressed with 6350432 - https://github.com/nuxeo/nuxeo-ai/compare/7f01bfb1280f48e707a1ad6e538842fe508cc602..63504328f34dade871510773dd4e9c18e567a456

nuxeo-ai-model/src/main/java/org/nuxeo/ai/cloud/NuxeoCloudClient.java
jjliggett: This has been addressed with 6350432 - https://github.com/nuxeo/nuxeo-ai/compare/7f01bfb1280f48e707a1ad6e538842fe508cc602..63504328f34dade871510773dd4e9c18e567a456

andreinechaev (approved):

nuxeo-ai-model/src/main/java/org/nuxeo/ai/cloud/NuxeoCloudClient.java
nichiporenko: Looks good! Thanks for the update

nuxeo-ai-model/src/main/java/org/nuxeo/ai/cloud/NuxeoCloudClient.java
nichiporenko: Looks good! Thanks for the update

nichiporenko (approved):

PR #639, opened 2025-05-27
Fix/failsafe dependency upgrade fix test case

debjit-hyland:
Testing PR builds to fix previous failure issues in - https://github.com/nuxeo/nuxeo-ai/pull/637


pom.xml
```diff
@@ -4,7 +4,7 @@
   <parent>
     <groupId>org.nuxeo</groupId>
     <artifactId>nuxeo-parent</artifactId>
-    <version>2023.0</version>
+    <version>2023.16</version>
```
troger: While we're at upgrading, I think you could align here on the latest 2023 version, `2023.30`

debjit-hyland: done

debjit-hyland:
Things I want to achieve through this PR - 

- Bump Nuxeo LTS version to LTS 2023 latest (2023.31)
- Bump build workflow to run on Node@20
- Remove org.tensorflow.proto dependency from ai-core pom.xml module as it is deprecated and having java classes generated from old 3.5.1 protobuf.
- Include com.google.protobuf latest version in parent pom.xml
- Use .proto files from tensorflow to generate latest java files using protoc 29.3. This mitigates DOS vulnerability which was reported in the build warnings - https://github.com/protocolbuffers/protobuf/security/advisories/GHSA-h4h5-3hr4-j3g2
- Add readme.md file with steps to generate proto files in future using latest protoc.

FYI - org.tensorflow.proto dependency was removed as it is not maintained anymore and the last version(1.15.0) had vulnerabilities reported due to usage of old .java classes generated internally using protobuf - 3.5.1 (See -  https://github.com/protocolbuffers/protobuf/security/advisories/GHSA-h4h5-3hr4-j3g2). 

I have included the proto files in the codebase which is used to generate the latest java files using latest protoc. This approach is chosen keeping in mind the usecase of the Protobuf message types used in the codebase remains as is.

addons/nuxeo-ai-aws-core/src/main/java/org/nuxeo/ai/transcribe/AudioTranscription.java
debjit-hyland, line 1: This will be removed on final commit before merge. Pls ignore as of now _/\_

nuxeo-ai-pipes/src/main/java/org/nuxeo/ai/pipes/functions/PropertyUtils.java
```diff
@@ -126,7 +126,7 @@ public static <T> T getPropertyValue(DocumentModel doc, String propertyName, Cla
      */
     public static String base64EncodeBlob(Blob blob) {
         try {
-            if (blob != null) {
+            if (blob != null && blob.getByteArray() != null && blob.getByteArray().length > 0) {
                 return Base64.encodeBase64String(blob.getByteArray());
             }
```
kevinleturc: An `InputStream` is created/fetched for each call to `getByArray`, this may be not performant to call it multiple times (depending on the blob implementation).
You should store the result:
```suggestion
            if (blob != null) {
                byte[] blobBytes = blob.getByteArray();
                 if (ArrayUtils.isNotEmpty(blobBytes)) {
                    return Base64.encodeBase64String(blobBytes);
                }
             }
```

nuxeo-ai-core/src/main/java/org/nuxeo/ai/enrichment/EnrichingStreamProcessor.java
```diff
@@ -196,20 +202,35 @@ protected Collection<? extends AIMetadata> cacheGet(String cacheKey) {
...
+                                                                            log.debug("Enrichment result is " + r);
+                                                                        }
+                                                                    })
+                                                                    .onFailure(failure -> {
+                                                                        //this is final failure after all retries
+                                                                        metrics.error();
+                                                                        log.warn("Enrichment error ({}) for record: {}", enricherName, record, failure);
+                                                                    });
+
+            return executor.get(supplier);
```
kevinleturc:
```suggestion
            return Failsafe.with(retryPolicy, circuitBreaker, timeout)
                                                                    .onSuccess(r -> {
                                                                        metrics.success();
                                                                        if (log.isDebugEnabled()) {
                                                                            log.debug("Enrichment result is " + r);
                                                                        }
                                                                    })
                                                                    .onFailure(failure -> {
                                                                        //this is final failure after all retries
                                                                        metrics.error();
                                                                        log.warn("Enrichment error ({}) for record: {}", enricherName, record, failure);
                                                                    })
                                                                    .get(supplier);
```

ataillefer (commented):
Sounds good globally. Two points though:
- I assume that the commented code and comments about testing will be removed in the end?
- About the generated code, could it be genereated during the build pipeline and put inside a dedicated folder (e.g.: `target/generated-sources`), ignored by Git? Maybe using the Maven `generate-sources`phase.

nuxeo-ai-core/src/main/java/org/nuxeo/ai/enrichment/EnrichingStreamProcessor.java
```diff
@@ -196,20 +202,35 @@ protected Collection<? extends AIMetadata> cacheGet(String cacheKey) {
...
+                log.warn("Enrichment error ({}) for record: {}", enricherName, record, ((ExecutionAttemptedEvent) c).getLastFailure());
+
+            });
+
+            FailsafeExecutor<Collection<AIMetadata>> executor = Failsafe.with(retryPolicy, circuitBreaker, timeout)
+                                                                    .onSuccess(r -> {
+                                                                        metrics.success();
+                                                                        if (log.isDebugEnabled()) {
+                                                                            log.debug("Enrichment result is " + r);
+                                                                        }
```
guirenard:
```suggestion
                                                                        log.debug("Enrichment result is {}", r);
```

nuxeo-ai-core/src/main/java/org/nuxeo/ai/tensorflow/TFRecord.java
```diff
@@ -17,30 +17,25 @@
...
-import com.google.protobuf.MessageOrBuilder;
-import com.google.protobuf.Parser;
-import com.google.protobuf.UnknownFieldSet;
+
 
 /**
  * Partial implementation of a Protobuf object that represents Tensorflow record
  */
-public class TFRecord extends GeneratedMessageV3 implements MessageOrBuilder {
+
```
guirenard: a bunch of emtpy lines in this class, we don't have empty lines between method and javadoc

nuxeo-ai-core/src/main/java/org/nuxeo/ai/tensorflow/TFRecord.java
```diff
@@ -73,11 +68,13 @@ public TFRecord(String docId, Features features) {
         this.features = features;
     }
 
-    /**
+
+/**
```
guirenard: bad indent

nuxeo-ai-core/src/main/java/org/nuxeo/ai/tensorflow/TFRecord.java
```diff
@@ -176,13 +173,15 @@ public int getSerializedSize() {
         return size;
     }
 
-    /**
+
+/**
```
guirenard: bad indent

nuxeo-ai-core/src/main/java/org/nuxeo/ai/enrichment/EnrichingStreamProcessor.java
debjit-hyland: Reverted the inline Failsafe.with(...).get(supplier) change due to a type inference issue when using multiple policies. The generic type <Collection<AIMetadata>> causes a compilation error (Wrong number of type arguments: 1; required: 2) because the overloaded Failsafe.with(...) method expects two type parameters in this context.

Keeping the original version with an explicitly typed FailsafeExecutor<Collection<AIMetadata>> avoids ambiguity and ensures type safety without requiring additional casting or verbose generics.

kevinleturc: Closing this comment and opening a new one on the same section.

nuxeo-ai-core/src/main/java/org/nuxeo/ai/enrichment/EnrichmentProvider.java
```diff
@@ -56,19 +61,19 @@ public interface EnrichmentProvider {
      */
     @SuppressWarnings("unchecked")
     default RetryPolicy getRetryPolicy() {
```
kevinleturc: Types should now be given to `RetryPolicy`:
```suggestion
    default RetryPolicy<Collection<AIMetadata>> getRetryPolicy() {
```
Ideally all implementation should receive type too.

nuxeo-ai-core/src/main/java/org/nuxeo/ai/enrichment/EnrichmentProvider.java
```diff
@@ -56,19 +61,19 @@ public interface EnrichmentProvider {
...
+        return new RetryPolicy<Collection<AIMetadata>>()
+            .abortOn(NuxeoException.class, FatalEnrichmentError.class)
+            .withBackoff(3,36, ChronoUnit.SECONDS)
+            .withMaxRetries(2);
     }
 
     /**
      * The circuit breaker for the provider
      */
     default CircuitBreaker getCircuitBreaker() {
```
kevinleturc: Same thing than getRetryPolicy:
```suggestion
    default CircuitBreaker<Collection<AIMetadata>> getCircuitBreaker() {
```

nuxeo-ai-core/src/main/java/org/nuxeo/ai/enrichment/EnrichingStreamProcessor.java
kevinleturc, line 1: `retryPolicy` and `circuitBreaker` attributes should receive their types:
```
        protected RetryPolicy<Collection<AIMetadata>> retryPolicy;

        protected CircuitBreaker<Collection<AIMetadata>> circuitBreaker;
```

nuxeo-ai-core/src/main/java/org/nuxeo/ai/enrichment/EnrichingStreamProcessor.java
```diff
@@ -196,20 +202,31 @@ protected Collection<? extends AIMetadata> cacheGet(String cacheKey) {
...
-                }
-            }).onFailedAttempt(failure -> {
-                metrics.error();
-                log.warn("Enrichment error ({}) for record: {} ", enricherName, record, failure);
-            }).onRetry(c -> {
+
+            CheckedSupplier<Collection<AIMetadata>> supplier = callable::call;
+            Timeout<Collection<AIMetadata>> timeout = Timeout.of(Duration.ofSeconds(60));
+
+            retryPolicy.onRetry(c -> {
```
kevinleturc: I'm not sure this is what we want, as I understand it the given listener will be registered to the `retryPolicy` and not the ongoing failsafe execution, the retry policy should be copied.

See next comment.

nuxeo-ai-core/src/main/java/org/nuxeo/ai/enrichment/EnrichingStreamProcessor.java
```diff
@@ -196,20 +202,31 @@ protected Collection<? extends AIMetadata> cacheGet(String cacheKey) {
...
+                        metrics.success();
+                        log.debug("Enrichment result is: {}", r);
+                    })
+                    .onFailure(failure -> {
+                        //this is final failure after all retries
+                        metrics.error();
+                        log.warn("Enrichment error ({}) for record: {}", enricherName, record, failure);
+                    });
+
+            return executor.get(supplier);
```
kevinleturc: When types are given to methods, attributes and others, the whole method could be re-wrote as:
```
        protected Collection<AIMetadata> callProvider(Record record, Callable<Collection<AIMetadata>> callable) {

            Timeout<Collection<AIMetadata>> timeout = Timeout.of(Duration.ofSeconds(60));

            var retryPolicy = this.retryPolicy.copy().onRetry(c -> {
                metrics.retry();
                log.debug("Retrying record: {}", record);
                //If we are retrying it means it has errorred out so metric has to be taken care of in higher versions of failsafe
                metrics.error();
                log.warn("Enrichment error ({}) for record: {}", enricherName, record, c.getLastFailure());

            });

            return Failsafe.with(retryPolicy, circuitBreaker, timeout)
                    .onSuccess(r -> {
                        metrics.success();
                        log.debug("Enrichment result is: {}", r);
                    })
                    .onFailure(failure -> {
                        //this is final failure after all retries
                        metrics.error();
                        log.warn("Enrichment error ({}) for record: {}", enricherName, record, failure.getFailure());
                    }).get(callable::call);
        }
```

I successfully compiled the module.

Note: There's some fixes in the suggestion like the warn log in the onFailure listener which now receive the exception from the failure.

nuxeo-ai-core/pom.xml
```diff
@@ -10,10 +10,10 @@
   <name>Ai core</name>
   <description>Core AI functionality</description>
   <dependencies>
-    <dependency>
-      <groupId>org.tensorflow</groupId>
-      <artifactId>proto</artifactId>
-    </dependency>
+      <dependency>
+          <groupId>com.google.protobuf</groupId>
```
guirenard: bad indent

You probably want to have a look to https://hyland.atlassian.net/wiki/spaces/NuxEng/pages/2232485639/Code+Formatting

PR #728, opened 2026-03-12
NXCON-228: Fix target platform version in nuxeo-ai packages to support all 2025.x releases

abhishek-gupta-nuxeo:


debjit-hyland (approved):

shristyPriyadarshini (approved):

addons/nuxeo-ai-aws-package/src/main/resources/package.xml
```diff
@@ -15,10 +15,10 @@
   <!-- Production State: production_ready | testing | proto -->
   <production-state>production_ready</production-state>
   <supported>true</supported>
-  <platforms>
-    <platform>@DISTRIBUTION_NAME@-@NUXEO_PLATFORM_VERSION@</platform>
-    <platform>@DISTRIBUTION_NAME@-@NUXEO_TARGET_VERSION@</platform>
-  </platforms>
+  <target-platform>
+    <name>@DISTRIBUTION_NAME@</name>
+    <version>@NUXEO_TARGET_VERSION@</version>
```
debjit-hyland: Shall we use [@NUXEO_PLATFORM_VERSION@,2026.0) instead as it is a platform-level property that is inherited automatically @kevinleturc ?

kevinleturc (commented):
The target platform should be deduced from the nuxeo-parent pom, see for example other packages: https://github.com/nuxeo/nuxeo-coldstorage/blob/lts-2025/nuxeo-coldstorage-package/src/main/resources/package.xml#L9-L12

This is mainly because we, as the team maintaining the server, ensure a backward compatibility (ie: something built against 2025.2 will work on 2025.3) but not a frontward compatibility (ie: something built against 2025.3 won't necessarily work on 2025.2).

About addon and server compatibility, our first wish as Nuxeo Eng Teams is to always depend on the latest released Nuxeo version, which should not be an issue as customers should always update to the latest version when available. But, you can still relax this if this is a wish or a need, and for that you should still use the version retrieval mechanism present in other addons + set the nuxeo-parent version to `2025.0`.

From my opinion, I think you can put in place/update the version retrieval mechanism + set the nuxeo-parent version to `2025.0`, do a release in order to have one version of the addon compatible with every 2025.x version, and then go back to the latest nuxeo-parent version.
If you go this way, you'll need to disable the automatic nuxeo-parent update the time you will downgrade the nuxeo-parent version, do the release and then re-enable the automatic update. This automatic update mechanism is here: https://github.com/nuxeo/nuxeo-ai/blob/2023/.github/workflows/_default_branch-update-nuxeo-parent.yml#L28-L29

addons/nuxeo-ai-aws-package/src/main/resources/package.xml
kevinleturc: Yes, you should align on what is done on other addons, see: https://github.com/nuxeo/nuxeo-ai/pull/728#pullrequestreview-3952039615.

kevinleturc (approved):

